PR #2, opened 2013-03-19
Fixed multi-thread connection initialization for ZPsycopgDA

zagy:
The connection initialization (transaction isolation, typecasts) was only done for first connection. When there are multiple threads running in parallel, connections where used which had not been initialized correctly.


distributist:
@nboullis (and anyone else who cares): I just finished upgrading an old Zope application from Python Zope 2.10 and Python 2.4.6 to Zope 2.13 and Python 2.7.9. I've just run into what looks like this bug under certain circumstances (and this explains why it only happens in production and not my development/test environments). We're using psycopg2 2.5.3 and ZPsycopgDA 2.4.6. 

Do you have any feedback on actually using your patch, or any others? I'm reluctant to patch something as low-level as this without some idea of what others have found.

And yes, replacing all the ZSQL with something more modern would be ideal; but it would be a huge job and my client ain't about to pay for it. ;-)

Thanks,
Hugh Ranalli

zagy:
@distributist: my client is using it in production 2013 now.

hranalli:
@zagy,
Thanks for the feedback. We've been doing some testing and have applied @nboullis patch, as it seemed the simpler approach. After a couple of weeks in production, with easily over 500K queries per day, we've seen no problems. The situation that in particular caused this problem to rear its ugly head has not recurred since applying the patch.  

Hugh

PR #16, opened 2014-05-25
Feature: cursor.callproc could accept Dict for PostgreSQL 9.0+ named parameters

mrmilosz:
Starting with PostgreSQL 9.0, functions can be called using "named parameters": https://wiki.postgresql.org/wiki/What's_new_in_PostgreSQL_9.0#Named_Parameter_Calls

This is especially useful for functions with a large number of arguments, as it allows you to refer to the arguments by name instead of having to remember their order. This small patch allows cursor.callproc to accept a Dict instead of a List to mirror PostgreSQL's feature. No previous functionality has been removed. 

Like the name of the stored procedure itself, the names of the parameters are NOT sanitized. This means that, just as the programmer should not expose control of the name of the stored procedure to the end user, the programmer should not expose control of the parameter names either.

A macro was added to python.h for the python-version-agnostic conversion of text objects into C char arrays (specifically the procedure parameter names, which come from the Dict keys.)

The code compiles and runs correctly under Python 2.7 and Python 3. There were no existing test suites for callproc (that I could find) so I wasn't sure where to put test suites for this new functionality. I used the following code to test:

https://gist.github.com/mrmilosz/905a4492216052821c38

Any procedure returning a value and matching the prototype will work for the test.

Thank you for your time. I hope you can make this patch a part of the trunk as it introduces a feature I would very much like to use! Of course, if I made any silly mistakes you are free to fix them first. For what it's worth I tried to pay close attention to my new/borrowed references and Py_DECREFs, but I am new to the Python C API.

### DIFF
--- a/doc/src/cursor.rst
+++ b/doc/src/cursor.rst
@@ -201,12 +201,17 @@ The ``cursor`` class
             
         Call a stored database procedure with the given name. The sequence of
         parameters must contain one entry for each argument that the procedure
-        expects. The result of the call is returned as modified copy of the
-        input sequence. Input parameters are left untouched, output and
-        input/output parameters replaced with possibly new values.
-        
-        The procedure may also provide a result set as output. This must then
-        be made available through the standard |fetch*|_ methods.
+        expects. Overloaded procedures are supported. Named parameters can be
+        used with a PostgreSQL 9.0+ client by supplying the sequence of
+        parameters as a Dict.
+
+        This function is, at present, not DBAPI-compliant. The return value is
+        supposed to consist of the sequence of parameters with modified output
+        and input/output parameters. In future versions, the DBAPI-compliant
+        return value may be implemented, but for now the function returns None.
+
+        The procedure may provide a result set as output. This is then made
+        available through the standard |fetch*|_ methods.
 
 
     .. method:: mogrify(operation [, parameters])
--- a/psycopg/cursor_type.c
+++ b/psycopg/cursor_type.c
@@ -1019,6 +1019,7 @@ psyco_curs_callproc(cursorObject *self, PyObject *args)
     int using_dict;
 #if PG_VERSION_HEX >= 0x090000
     PyObject *pname = NULL;
+    PyObject *spname = NULL;
     PyObject *bpname = NULL;
     PyObject *pnames = NULL;
     char *cpname = NULL;
@@ -1049,7 +1050,7 @@ psyco_curs_callproc(cursorObject *self, PyObject *args)
 
     using_dict = nparameters > 0 && PyDict_Check(parameters);
 
-    /* A Dict is complicated. The parameter names go into the query */
+    /* a Dict is complicated; the parameter names go into the query */
     if (using_dict) {
 #if PG_VERSION_HEX >= 0x090000
         if (!(pnames = PyDict_Keys(parameters))) {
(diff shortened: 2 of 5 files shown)


mrmilosz:
Hi Daniele,

I apologize for dropping the ball on this! My travels ended up quite long, and when I returned I started a new job and moved cities. Named parameters in psycopg2 fell victim to MRU.

After over a year I've started again to work on the project that brought upon my need for named parameters in psycopg2, and, while deploying the project to my dev box, ran into "TypeError: 'dict' object does not support indexing". My memories of working on this feature came rushing back.

Is there still any interest in continuing this endeavour? At least from my end, there certainly is. Unless, that is, someone else has been working on something similar, or something that obsoletes this change?

mrmilosz:
Merged in your named-callproc branch then merged in master after 1.5 years. Everything worked mostly without conflict. Did have to change PG_VERSION_HEX to PG_VERSION_NUM. That really tripped me up, but I was rescued by grep :)

mrmilosz:
OK, reading over the correspondence here it seems the next step is to figure out what to do with the DBAPI spec. I'm going to try to convince the ML that returning anything at all from callproc is redundant, as everything is available via returned rows anyway, where it has dictionary-like access if you need it.

mrmilosz:
https://mail.python.org/pipermail/db-sig/2015-December/006193.html

I've been using callproc heavily on my branch (now merged with mainline) for over a week now, and it's quite stable. Reading over your callouts, it would appear you've implemented all these changes.

After some discussion on the forum, I still think simply returning Py_None is the best compromise here, but for backwards-compatibility I am content with res = parameters. It doesn't make a difference in my use-case.

Shall I make that so?

dvarrazzo:
I need a bit of concentration to get back to this thread: I had it completely swapped off my brain :) I barely followed the dbsig conversation and not really grokked the differences between postgres and other dbs. AFAICS psycopg has never used the OUT parameters in the return value but always returned the input argument, likely always ignored by the caller.

```
In [8]: cur.execute("create function double(n int, out n2 int) language sql as 'select 2 * $1'")
In [10]: cur.callproc("double", [10])
Out[10]: [10]
In [11]: cur.fetchone()
Out[11]: (20,)
```

For me it is fine to maintain the behaviour and keep returning output = input and only rely on the `fetch*` to get the 0, 1, n returned tuples. If it makes the code easier we may also return only the values in the positional parameters and not the named one: I'm fine just ignoring the not particularly meaningful dbapi requirement as long as for pre-existing code (callproc called with positional only arguments) the return value isn't changed.

dvarrazzo:
It would be better if you gave a name to your branch and rebased it on master.

dvarrazzo:
Also note that in master there is now another use of `PQescapeIdentifier`: it would be worth refactoring dropping the ad-hoc wrapper used for the named params and use only one.

mrmilosz:
OK, I will:

* Make the call return the parameter tuple when invoked using a tuple
* Keep it returning None when invoked using a dictionary
* Consolidate uses of PQescapeIdentifier

In terms of the branch, right now my workflow involves regularly rebasing my master branch off the psycopg2 master branch. What is the value of naming my branch something other than master? Does that affect the pull request?

mrmilosz:
Oh, are you saying to cancel this pull request and rebase my branch off the current master to avoid the giant merge? I can do that.

mrmilosz:
Closing this pull request and continuing work using a new branch called named-callproc which will be renamed to named-callproc.

New pull request: https://github.com/psycopg/psycopg2/pull/381

PR #17, opened 2014-06-20
Add context manager for connection pool for use with the with statement

irrelative:
Hi devs!

I really enjoy using context managers for connections and cursors, but it's kind of a pain to do this with connection pooling. This adds the ability to do this, and works like so:

```
pool = SimpleConnectionPool(...)

def work():
    with pool() as conn:
        # Use connection normally, it'll automatically be put back into the pool when complete
```

I'm open to suggestions with the interface -- the `__call__` seemed the least contentious with the existing method names. Let me know what you think.

Cheers,
Justin

### DIFF
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -32,6 +32,20 @@ class PoolError(psycopg2.Error):
     pass
 
 
+class ConnectionContext(object):
+
+    def __init__(self, pool):
+        self.pool = pool
+        self._conn = None
+
+    def __enter__(self):
+        self._conn = self.pool.getconn()
+        return self._conn
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.pool.putconn(self._conn)
+
+
 class AbstractConnectionPool(object):
     """Generic key-based pooling code."""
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
I wanted to include this for 2.6, but actually took a look and the solution/design and I don't like it.

`pool()` returns a conn in a context, but then that connection doesn't behave like the connection's context. In the connection context the transaction is committed/rolled back on exit, whereas on the pool it is put back, so it gets rolled back if a transaction is active, even in case of no error.

I also don't think using `__call__` is required. If `__enter__` and `__exit__` are on the pool itself I think you can have:

    with pool as conn:
        ...

(not that I necessarily like that: I'm just talking from a syntactical PoV).

So, I'm happy to think about a solution. I think the easiest thing to put together is something that would get used as:

    with pool as conn:
        with conn:
            with conn.cursor() as cur:
                # do some work
            # on exit the cursor is closed
        # on exit the transaction is terminated
    # on exit the connection is returned to the pool

which seems awkward... can we do better?

irrelative:
After looking at this again, I tend to agree with you. I was trying to avoid another indentation block if possible, but it does seem to make the implementation less clear. I'll try more straightforward approach, and post the diff in a bit.

PR #265, opened 2014-10-01
Allow pickling of cursor.description

ojrac:
This is for people using dtuple.py; a dtuple.DatabaseTuple instance keeps a reference to cursor.description, which is not picklable because psycopg2 doesn't export the Column namedtuple it uses.

This commit exports the Column namedtuple, and includes a test to verify the pickle/unpickle works after exporting Column.


dvarrazzo:
Merged into my branch, will end up in the repos as soon as the tests are finished.

PR #309, opened 2015-04-23
fix MinTimeLoggingCursor.callproc

djsfcom:


dvarrazzo:
To be merged with a more meaningful message and a note in the newsfile

dvarrazzo:
Merged

PR #320, opened 2015-05-21
Notify example should pop oldest message in conn.notifies, not newest

btubbs:
The example in the docs of handling LISTEN/NOTIFY does a conn.notifies.pop(), which will pull the most-recently-received notification, rather than the oldest un-popped notification, if more than one is waiting.  I think most users would prefer to receive notifications in the order they were sent, so I've updated the example to provide that behavior by doing conn.notifies.pop(0).

### DIFF
--- a/doc/src/advanced.rst
+++ b/doc/src/advanced.rst
@@ -291,7 +291,7 @@ something to read::
         else:
             conn.poll()
             while conn.notifies:
-                notify = conn.notifies.pop()
+                notify = conn.notifies.pop(0)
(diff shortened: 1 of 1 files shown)


dvarrazzo:
It is true, however pop(0) is less efficient than pop(). The matter of fact is that notifies should have been a deque, not a list. This will be changed in the future. Let me think a bit more about these docs...

dvarrazzo:
Merged. It's a good idea to document the idea in its simplest form, not a premature optimization.

dvarrazzo:
...and thank you ;)

Issue #35, opened 2011-01-17
Add support to libpq version query


dvarrazzo:
Can be for the version we compiled against and the runtime one since pg 9.1.

dvarrazzo:
Docs: http://www.postgresql.org/docs/9.1/static/libpq-misc.html#LIBPQ-PQLIBVERSION

we may expose a `__libpq_version__` constant to expose the version compiled in and a `extensions.libpq_version()` function to return the `PQlibVersion()` value. If the compiled version is < 9.1 this version should raise a NotSupported exception.

dvarrazzo:
Merged to master.

PR #321, opened 2015-06-01
Add parse_dsn module function

a1exsh:
Calls PQconninfoParse to parse the dsn into a list of keyword and value
structs, then constructs a dictionary from that.  Can be useful when one
needs to alter some part of the the connection string reliably, but
doesn't want to get into all the details of parsing a dsn string:
quoting, URL format, etc.

### DIFF
--- a/doc/src/module.rst
+++ b/doc/src/module.rst
@@ -78,6 +78,7 @@ The module interface respects the standard defined in the |DBAPI|_.
 
     .. seealso::
 
+        - `parse_dsn`
         - libpq `connection string syntax`__
         - libpq supported `connection parameters`__
         - libpq supported `environment variables`__
@@ -91,6 +92,17 @@ The module interface respects the standard defined in the |DBAPI|_.
         The parameters *connection_factory* and *async* are Psycopg extensions
         to the |DBAPI|.
 
+.. function:: parse_dsn(dsn)
+
+    Parse connection string into a dictionary of keywords and values.
+
+    Uses libpq's ``PQconninfoParse`` to parse the string according to
+    accepted format(s) and check for supported keywords.
+
+    Example::
+
+        >>> psycopg2.parse_dsn('dbname=test user=postgres password=secret')
+        {'password': 'secret', 'user': 'postgres', 'dbname': 'test'}
 
 .. data:: apilevel
 
--- a/doc/src/module.rst
+++ b/doc/src/module.rst
@@ -78,7 +78,7 @@ The module interface respects the standard defined in the |DBAPI|_.
 
     .. seealso::
 
-        - `parse_dsn`
+        - `~psycopg2.extensions.parse_dsn`
         - libpq `connection string syntax`__
         - libpq supported `connection parameters`__
         - libpq supported `environment variables`__
@@ -91,18 +91,6 @@ The module interface respects the standard defined in the |DBAPI|_.
 
         The parameters *connection_factory* and *async* are Psycopg extensions
         to the |DBAPI|.
-
-.. function:: parse_dsn(dsn)
-
-    Parse connection string into a dictionary of keywords and values.
-
-    Uses libpq's ``PQconninfoParse`` to parse the string according to
-    accepted format(s) and check for supported keywords.
-
-    Example::
-
-        >>> psycopg2.parse_dsn('dbname=test user=postgres password=secret')
-        {'password': 'secret', 'user': 'postgres', 'dbname': 'test'}
 
 .. data:: apilevel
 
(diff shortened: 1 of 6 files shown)


doc/src/module.rst
```diff
@@ -91,6 +92,17 @@ The module interface respects the standard defined in the |DBAPI|_.
         The parameters *connection_factory* and *async* are Psycopg extensions
         to the |DBAPI|.
 
+.. function:: parse_dsn(dsn)
+
```
dvarrazzo: This function should be exposed by psycopg2.extensions, not the main module.

psycopg/psycopgmodule.c
```diff
@@ -112,6 +112,44 @@ psyco_connect(PyObject *self, PyObject *args, PyObject *keywds)
...
+    PQconninfoOption *options = NULL, *o;
+    PyObject *res = NULL, *value;
+
+    if (!PyArg_ParseTuple(args, "s", &dsn)) {
+        return NULL;
+    }
+
+    options = PQconninfoParse(dsn, &err);
+    if (!options) {
+        PyErr_Format(PyExc_RuntimeError, "PQconninfoParse: %s: %s", dsn, err);
```
dvarrazzo: This error is wrong. The dsn shouldn't be printed back as it may contain passwords that may end up in the logs. It shoud be a ProgrammingError if the dsn is wrong (with a human error message such as "error parsing the dsn: %s" % err) or an OperationalError if err is null as that's a memory allocation error according to the docs).

psycopg/psycopgmodule.c
```diff
@@ -112,6 +112,44 @@ psyco_connect(PyObject *self, PyObject *args, PyObject *keywds)
...
+        PQfreemem(err);
+        return NULL;
+    }
+
+    res = PyDict_New();
+    for (o = options; o->keyword != NULL; o++) {
+        if (o->val != NULL) {
+            value = PyString_FromString(o->val);
+            if (value == NULL || PyDict_SetItemString(res, o->keyword, value) != 0) {
```
dvarrazzo: This is terrible, separate the error check from PyString_FromString from the SetItemString!

It is also a reference leak: SetItemString doesn't steal the reference. value must be decref'd.

psycopg/psycopgmodule.c
```diff
@@ -112,6 +112,44 @@ psyco_connect(PyObject *self, PyObject *args, PyObject *keywds)
     return conn;
 }
 
+#define psyco_parse_dsn_doc \
+"parse_dsn(dsn) -- Parse database connection string.\n\n"
```
dvarrazzo: better: parse_dsn(dsn) -> dict. Also I don't think you need the \n's. Even better put all the documentation here and use autodoc to get it in the docs.

psycopg/psycopgmodule.c
```diff
@@ -112,6 +112,44 @@ psyco_connect(PyObject *self, PyObject *args, PyObject *keywds)
...
+    if (!options) {
+        PyErr_Format(PyExc_RuntimeError, "PQconninfoParse: %s: %s", dsn, err);
+        PQfreemem(err);
+        return NULL;
+    }
+
+    res = PyDict_New();
+    for (o = options; o->keyword != NULL; o++) {
+        if (o->val != NULL) {
+            value = PyString_FromString(o->val);
```
dvarrazzo: Does it work ok on Python3? Probably you need a function like Text_FromUTF8 -- see psycopg/python.h and example of usage all around the lib.

dvarrazzo:
Thank you for the contribution.

The patch needs tests, both failing and successful, and possibly with some tricky values (quotes, non-quotes, spaces etc).

a1exsh:
Excellent points, thank you!  I will have another look on it and re-submit.

a1exsh:
I've addressed your feedback on this patch, please have a look when you got a minute.

dvarrazzo:
Thank you, it seems much better.

a1exsh:
Good catch with libq version.  I think it's just a matter of exposing PQlibVersion in python and having @skip_before_libpq decorator?

dvarrazzo:
Exposing the libpq version is another thing we should have done for a long time, see issue #35.

We sort of dragged it because at the time of opening the likelyhood of that function not being available was high, but now we are planning several features for psycopg 2.7 requiring a more modern libpq so I guess it's time to implement it.

a1exsh:
Hey Daniele, could you please merge this?  We could really use reliable DSN parsing in #322 in order to specify the `replication=true|database` correctly.  Thanks!

dvarrazzo:
Ok, I'll give it a final review and merge it.

dvarrazzo:
I've made a few fixes and merged, thank you very much.

PR #322, opened 2015-06-01
Add support for streaming replication protocol

a1exsh:
Introduce ReplicationConnection and ReplicationCursor classes, that
incapsulate initiation of special type of PostgreSQL connection and
handling of special replication commands only available in this special
connection mode.

The handling of stream of replication data from the server is modelled
largely after the existing support for "COPY table TO file" command and
pg_recvlogical tool supplied with PostgreSQL (though, it can also be
used for physical replication.)


lib/extras.py
```diff
@@ -437,6 +437,144 @@ def callproc(self, procname, vars=None):
...
+class ReplicationConnection(_connection):
+    """A connection that uses `ReplicationCursor` automatically."""
+
+    def __init__(self, *args, **kwargs):
+        """Initializes a replication connection, by adding appropriate replication parameter to the provided dsn arguments."""
+
+        if len(args):
+            dsn = args[0]
+
+            # FIXME: could really use parse_dsn here
```
dvarrazzo: Agreed

psycopg/pqpath.c
```diff
@@ -40,7 +40,14 @@
 #include "psycopg/pgtypes.h"
 #include "psycopg/error.h"
 
+#include "postgres_fe.h"
+#include "access/xlog_internal.h"
+#include "common/fe_memutils.h"
```
dvarrazzo: These two includes I'm not sure are in libpq-dev

psycopg/pqpath.c
```diff
@@ -40,7 +40,14 @@
...
+#include "libpq-fe.h"
+
 #include <string.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
```
dvarrazzo: Is this going to work on Windows?

dvarrazzo:
I'd like to look better into this patch: I don't have much experience about replication.

I guess a test in the test suite would be hard to fit, but it'd be nice to have at least an example script.

lib/extras.py
```diff
@@ -437,6 +437,144 @@ def callproc(self, procname, vars=None):
...
+
+    def identify_system(self):
+        """Get information about the cluster status."""
+
+        self.execute("IDENTIFY_SYSTEM")
+        return dict(zip(['systemid', 'timeline', 'xlogpos', 'dbname'],
+                        self.fetchall()[0]))
+
+    def quote_ident(self, ident):
+        # FIXME: use PQescapeIdentifier or psycopg_escape_identifier_easy, somehow
```
dvarrazzo: When this will be released we would have PQescapeIdentifier exposed, so yes, it will happen.

dvarrazzo:
It would be cool if you wanted to have a chat on the ML about what this new feature would provide. Is it something you are already using in production?

a1exsh:
This is experimental code, it is not yet used in production.

The main goal is to provide a python client for logical decoding stream of WAL log entries, a feature that has been added in 9.4.

psycopg/pqpath.c
a1exsh: Oh, apparently I wanted to push this a bit too much :-p  I've committed a fix, it requires copying a number of postgres-internal typedefs and const, though.

psycopg/pqpath.c
a1exsh: Excellent point, addressed in a new commit.

a1exsh:
Speaking of example code, this is what I'm currently using only for testing:

    from io import TextIOBase
    import sys
    import psycopg2
    from psycopg2.extras import ReplicationConnection, REPLICATION_LOGICAL

    class LogicalWriter(TextIOBase): #(object)
        def write(self, buf):
            sys.stdout.write(buf)
            sys.stdout.write("\n")
            return True

    writer = LogicalWriter()

    conn = psycopg2.connect(dsn, connection_factory=ReplicationConnection)
    cur = conn.cursor()

    #cur.create_replication_slot(REPLICATION_LOGICAL, slot, "bottledwater")
    cur.start_replication(writer, REPLICATION_LOGICAL, slot_name=slot,
                          options={'format': 'json'})

On the server side we use a fork of bottledwater, compiled with JSON output support: https://github.com/a1exsh/bottledwater-pg/tree/feature/json-smallest-workable

What raises the most questions for us is the protocol for the Writer object to be passed to start_replication method.  I think it would be nice to pass some metadata to the write() callback that is available to the pqpath.c only at the moment, such as: values of dataStart, walEnd and sendTime fields from the message header, probably also written_lsn and fsync_lsn.

One can either pass these as positional parameters or keyword arguments, but that isn't very extensible, and it breaks the contract for sys.stdout (which is useful only for testing, but still useful).

I believe the most extensible way is to pass an object that can have attributes for the metadata, as well as methods to control reporting the fsync_lsn to the server, stopping replication (right now the only way for that is to throw an exception and catch it around start_replication call), and so on.  To make it compatible with sys.stdout usecase, one can even inherit the class from str, like this:

    class ReplicationMessage(str):
        #wal_end
        #data_start
        #send_time
        #def commit(self):
        #def stop(self):
        ...

So instead of cryptic "return True" the client would signal that it has stored the incoming message with msg.commit().  And if for some reason it needs to stop processing the stream, there is msg.stop().

dvarrazzo:
Hello,

thank you very much for the example. The problem is that this bug tracker is only monitored by few people: I'm sure the mailing list is a better place to discuss about this feature. Some times ago I spoke with some core developer, can't really remember who, and he indeed asked for COPY_BOTH support to play with replication. I couldn't work on the feature yet.

If you post on the ML and also reference the message to -hackers I'm sure you'll get some better look at your patch.

I'll review what you have made anyway but can't do it just now.

a1exsh:
OK, will do.

a1exsh:
Unfortunately, we cannot use copy_expert because the two-way communication via keepalive messages from client to the server is critical for proper implementation streaming replication.

One can consume data in the same way using copy_expert, but the server will never know how far the client is in terms of data stored reliably (fsync'ed).  As such, the server will retain the WAL files indefinitely, ultimately exhausting the disk space.  (Less annoying is the fact that if you stop and restart consuming the repl. stream, you will get the same stream of data again from the beginning, since the server was not told you don't need it anymore.)  There is a note about this in the doc part of this patch and it is explained in the Postgres streaming replication docs.

a1exsh:
Link to ML thread: http://www.postgresql.org/message-id/CACACo5STpV-T+4nNCjPNr2u=HnxnRTNVkas7Kf=uE90JZ1HA0w@mail.gmail.com

Ormod:
It would also be nice if there were a convenience function for streaming a basebackup tar a la pg_basebackup. (to a given file like object)

a1exsh:
So I've pushed here what I believe to be a mature version of this feature.

If accepted, the async handling approach may be also useful for async implementation of copy_from/to(), I hope.

a1exsh:
OK, after testing this patch at somewhat larger scale, a number of memory leaks were exposed, but I believe all of them are fixed now.  I do not expect to push more updates anytime soon, so please have another look when you have the time.

Thank you!

a1exsh:
Well, this is embarrassing, but I was so glad I've finally found that last mem leak, that I've also copied over GC-untrack call for "extra safety", but totally forgot to add the GC flag in the type definition.  Fixed now.

a1exsh:
And here we go again, yet another fix...

I guess a better strategy would be to get back to you when this code has seen production use on reasonable scale and duration.  Till then, sorry for the noise.

dvarrazzo:
Not at all: your "noise" is greatly appreciated :)

a1exsh:
Oh, thanks, that sort of a relief. :-)

ringerc:
I've been looking into this as part of support for logical decoding, since I want to use `psycopg2` for testing the protocol for a logical decoding output plugin. I'm pleased to see support already implemented.

Re using `copy_expert`, you can't do that because of #352. If `psycopg2` is patched to treat `PGRES_COPY_BOTH` as `PGRES_COPY_OUT` then it should work, but as @a1exsh noted that's not sufficient due to the need to support feedback. It's also unsuitable because for `COPY BOTH` mode you really *need* to process a stream of discrete messages; the length information in the `CopyData` message may be the only thing that makes a binary protocol stream parseable.

I outlined this in more detail on issue #351 .

I'll take a look at this PR and provide more detailed comments.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
 
 .. autoclass:: MinTimeLoggingCursor
 
+Replication cursor
+^^^^^^^^^^^^^^^^^^
+
+.. autoclass:: ReplicationConnection
+
+   This connection factory class can be used to open a special type of
+   connection that is used for streaming replication.
```
ringerc: Should note that normal SQL is not permitted or supported on a replication connection.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+   Example::
+
+       from psycopg2.extras import ReplicationConnection, REPLICATION_PHYSICAL, REPLICATION_LOGICAL
+       conn = psycopg2.connect(dsn, connection_factory=ReplicationConnection)
```
ringerc: Does the user need to supply `replication=database` or `replication=on` in the `dsn`? Or do you add that?

If connecting to 9.3 or older, which don't support `replication=database` and only do "physical" replication, how's that handled?

(I'm intentionally reading docs before code so I see from a user perspective and know what the docs don't cover).

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+    .. method:: start_replication(slot_type, slot_name=None, writer=None, start_lsn=None, timeline=0, keepalive_interval=10, options=None)
+
+       Start a replication stream.  On non-asynchronous connection, also
+       consume the stream messages.
+
+       :param slot_type: type of replication: either `REPLICATION_PHYSICAL` or
+                         `REPLICATION_LOGICAL`
+       :param slot_name: name of the replication slot to use (required for
+                         logical replication)
+       :param writer: a file-like object to write replication messages to
```
ringerc: This is IMO the wrong API here. A file-like object is not suitable for COPY BOTH, because you may need clearly delimited messages. Sometimes it's not possible to interpret a stream of binary results *without* knowing the message boundaries and lengths from `CopyData` messages.

IMO you should either have a listener pattern here, where you call an object that exposes a method like `copy_message(msgbytes)`, or you should return a generator that does something like yields byte strings (for each message) until generation end. The generator is strongly preferable IMO because it lets the client control the protocol flow and inject feedback messages as it goes.

I think that generator should be part of the base psycopg2 cursor and usable from `copy_expert` (or a variant) too, so that you can:

    for csvrow in curs.copy_expert('COPY myhugetable TO stdout', None):
        ... do something with the str object 'csvrow' ...

it's currently super-inefficient to have psycopg2 append the data to a file-like object, only to then parse it out again.

If we need a message-oriented API for COPY BOTH, which we do, why not support it for regular `copy` too?

I do not think this is suitable for merging with a file-like API and should be changed.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+       :param start_lsn: the LSN position to start from, in the form
+                         ``XXX/XXX`` (forward-slash separated pair of
+                         hexadecimals)
```
ringerc: That LSN is really just a uint64 formatted crazily for historical reasons. So an `int` should be acceptable too. See the implementation of the `pg_lsn` data type on the server side in 9.4+ for details and example code.

Needs extra text like:

> The LSN `0` (`'0/0'` in text format) may be passed to request that replay begins at the earliest available point.

and maybe:

> Note that replay may begin at a point later in history than the passed LSN if the data for that LSN is no longer present on the server. The client must ensure that repay has begun at the expected position using information in the COPY BOTH protocol stream.

... or just a warning to read the docs in detail there.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+       :param timeline: WAL history timeline to start streaming from (optional,
+                        can only be used with physical replication)
+       :param keepalive_interval: interval (in seconds) to send keepalive
```
ringerc: Should be documented as "synchronous mode only, ignored for asynchronous mode")? Or is it used? If so, how?

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+                                  messages to the server
+       :param options: a dictionary of options to pass to logical replication
```
ringerc: ... must be `None` for physical replication.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+       When used on non-asynchronous connection this method enters an endless
+       loop, reading messages from the server and passing them to ``write()``
+       method of the *writer* object.  This is similar to operation of the
+       `~cursor.copy_to()` method.  It also sends keepalive messages to the
+       server, in case there were no new data from it for the duration of
+       *keepalive_interval* seconds (this parameter's value must be equal to
+       at least than 1 second, but it can have a fractional part).
+
+       With asynchronous connection, this method returns immediately and the
+       calling code can start reading the replication messages in a loop.
```
ringerc: > start reading the replication messages in a loop

how?

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+       :param timeline: WAL history timeline to start streaming from (optional,
+                        can only be used with physical replication)
+       :param keepalive_interval: interval (in seconds) to send keepalive
+                                  messages to the server
+       :param options: a dictionary of options to pass to logical replication
+                       slot
+
+       When used on non-asynchronous connection this method enters an endless
+       loop, reading messages from the server and passing them to ``write()``
+       method of the *writer* object.  This is similar to operation of the
```
ringerc: What class is the message? That's not made clear here or in the sample Writer below

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+       First, like with the `~cursor.copy_to()` method, the code that calls
+       the provided ``write()`` method checks if the *writer* object is
+       inherited from `~io.TextIOBase`.  If that is the case, the message
+       payload to be passed is converted to unicode using the connection's
+       `~connection.encoding` information.  Otherwise, the message is passed
+       as is.
+
+       The *msg* object being passed is an instance of `~ReplicationMessage`
+       class.
```
ringerc: > The *msg* object being passed is an instance of `~ReplicationMessage` class.

This should be higher up, since it's confusing to read and understand the code above w/o knowing that.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+                   if self.should_report_to_the_server_now(msg):
+                       msg.cursor.send_replication_feedback(flush_lsn=msg.wal_end)
+
+       First, like with the `~cursor.copy_to()` method, the code that calls
+       the provided ``write()`` method checks if the *writer* object is
+       inherited from `~io.TextIOBase`.  If that is the case, the message
+       payload to be passed is converted to unicode using the connection's
+       `~connection.encoding` information.  Otherwise, the message is passed
+       as is.
```
ringerc: Very nice feature

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+    .. method:: start_replication(slot_type, slot_name=None, writer=None, start_lsn=None, timeline=0, keepalive_interval=10, options=None)
+
+       Start a replication stream.  On non-asynchronous connection, also
+       consume the stream messages.
+
+       :param slot_type: type of replication: either `REPLICATION_PHYSICAL` or
+                         `REPLICATION_LOGICAL`
+       :param slot_name: name of the replication slot to use (required for
+                         logical replication)
```
ringerc: For "physical" replication, no slot is required; it's permissible to replicate w/o a slot. It'd be nice to expose this in the API, e.g. by passing `None` as the slot name.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+       should send a confirmation message to the server.  This should be done
+       by calling `~send_replication_feedback()` method on the corresponding
+       replication cursor.  A reference to the cursor is provided in the
+       `~ReplicationMessage` as an attribute.
+
+       .. warning::
+
+           Failure to properly notify the server by constantly consuming and
+           reporting success at appropriate times can eventually lead to "disk
+           full" condition on the server, because the server retains all the
```
ringerc: Disk full is only a risk when using a replication slot, fwiw. Physical replication without a slot can't cause disk full, and is one of the reasons it remains useful.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+           On the other hand, it is not recommended to send a confirmation
+           after every processed message, since that will put an unnecessary
+           load on network and the server.  A possible strategy is to confirm
+           after every COMMIT message.
+
+    .. method:: stop_replication()
+
+        In non-asynchronous connection, when called from the ``write()``
+        method, tell the code in `~start_replication` to break out of the
+        endless loop and return.
```
ringerc: .. and in an async connection? Has no effect?

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+       :param slot_name: name of the replication slot to drop
+
+       Example::
+
+           cur.drop_replication_slot("testslot")
+
+    .. method:: start_replication(slot_type, slot_name=None, writer=None, start_lsn=None, timeline=0, keepalive_interval=10, options=None)
+
+       Start a replication stream.  On non-asynchronous connection, also
```
ringerc: What makes a connection synchronous or asynchronous? I presume you mean http://initd.org/psycopg/docs/advanced.html#asynchronous-support, i.e. it refers to psycopg2's built-in support for async connections. A cross-reference would be good here, along with making it a bit clearer how sync vs async works, which params are ignored for async if any, etc.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+        In case of asynchronous connection, if the feedback message cannot be
+        sent at the moment, remembers the passed LSN positions for a later
+        hopefully successful call or call to `~flush_replication_feedback()`.
+
+    .. method:: flush_replication_feedback(reply=False)
+
+        :param reply: request the server to send back a keepalive message immediately
+
+        This method tries to flush the latest replication feedback message
+        that `~send_replication_feedback()` was trying to send, if any.
```
ringerc: It "tries". What happens if it fails? How do we tell?

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+        :param decode: a flag indicating that unicode conversion should be
+                       performed on the data received from the server
+
+        This method should be used in a loop with asynchronous connections
+        after calling `~start_replication()` once.
+
+        It tries to read the next message from the server, without blocking
+        and returns an instance of `~ReplicationMessage` or *None*, in case
+        there are no more data messages from the server at the moment.
```
ringerc: Should make it clear that a whole message is returned; partial message data is buffered then returned once a complete message is available.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+        Such messages are silently consumed by this method and are never
+        reported to the caller.
+
+    .. method:: fileno()
+
+        Call the corresponding connection's `~connection.fileno()` method and
+        return the result.
+
+        This is a convenience method which allows replication cursor to be
+        used directly in `~select.select()` or `~select.poll()` calls.
```
ringerc: This will break on Windows.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+    .. attribute:: replication_io_timestamp
+
+        A `~datetime` object representing the timestamp at the moment of last
+        communication with the server (a data or keepalive message in either
+        direction).
+
+    An actual example of asynchronous operation might look like this::
```
ringerc: Excellent to see good examples like this in your docs. Very, very impressed.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+            timeout = keepalive_interval - (datetime.now() - cur.replication_io_timestamp).total_seconds()
+            if timeout > 0:
+                select.select([cur], [], [], timeout)
+
+.. autoclass:: ReplicationMessage
+
+    .. attribute:: payload
+
+        The actual data received from the server.  An instance of either
+        ``str`` or ``unicode``.
```
ringerc: Unclear how the user should expect to know which to expect.

For synchronous mode you describe that if the writer class extends io.TextIOBase the connection encoding is used to do a unicode conversion.

What about asynchronous mode? How do you get unicode conversion done for that? This needs an x-ref to a relevant section of docs or more detail.

doc/src/extras.rst
```diff
@@ -141,6 +141,262 @@ Logging cursor
...
+
+    .. attribute:: data_start
+
+        LSN position of the start of the message.
+
+    .. attribute:: wal_end
+
+        LSN position of the end of the message.
```
ringerc: I think you've misread the docs there.

See http://www.postgresql.org/docs/9.4/static/protocol-replication.html, message XLogData (B)

> Int64 The current end of WAL on the server.

The docs indicate that this message field is the *server's current xlog end position*, not the xlog position for the end of the message. Otherwise it'd be redundant anyway, since one message's end would be the next one's start.

lib/extras.py
ringerc: Yep, needs fixing before merge really

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+            if dsn.startswith('postgres://') or dsn.startswith('postgresql://'):
+                # poor man's url parsing
+                if dsn.rfind('?') > 0:
+                    if not dsn.endswith('?'):
+                        dsn += '&'
+                else:
+                    dsn += '?'
+            else:
+                dsn += ' '
+            dsn += 'replication=database'
```
ringerc: replication=database won't work on 9.3 or older, fwiw... but it's useful to have replication=true supported for 9.4+ too, because that lets you connect without knowing a particular database name when you plan to do streaming replication.

I suggest appending `replication=database` only if you don't already find a `replication` dsn term. That way the user can explicitly pass `replication=on` if they really want it, and it doesn't require any complicated API that's more effort than it's worth to add.

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+
+    def identify_system(self):
+        """Get information about the cluster status."""
+
+        self.execute("IDENTIFY_SYSTEM")
+        return dict(zip([_.name for _ in self.description],
+                        self.fetchall()[0]))
+
+    def quote_ident(self, ident):
+        # FIXME: use PQescapeIdentifier or psycopg_escape_identifier_easy, somehow
```
ringerc: Why's this exposed on the `ReplicationCursor`?

This is an oft-requested piece of functionality, but it should really be on the base psycopg2 Connection object IMO.

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+        return '"%s"' % ident.replace('"', '""')
+
+    def create_replication_slot(self, slot_type, slot_name, output_plugin=None):
+        """Create streaming replication slot."""
+
+        command = "CREATE_REPLICATION_SLOT %s " % self.quote_ident(slot_name)
```
ringerc: Looks like some more validation's needed here. Allowing `slot_name` to be None if physical replication is in use, but raising an exception if it's None for logical. Right now you'll get an error like 

    AttributeError: 'NoneType' object has no attribute 'replace'

which isn't very enlightening.

doc/src/extras.rst
a1exsh: First, thanks for your interest!

We add the `replication` keyword unless already supplied by the user.  Good point about 9.3 compatibility, I don't think I've ever tested that as my main focus was logical replication.  I will revisit this.

doc/src/extras.rst
a1exsh: Yes, it is about `psycopg2` async support.  The doc part of the patch is a rather messy on this, I agree.

doc/src/extras.rst
a1exsh: Right, should be added to the doc.

doc/src/extras.rst
a1exsh: For that there's an optional parameter `decode` in `cursor.read_replication_message()`.

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+
+        if slot_name:
+            command += "SLOT %s " % self.quote_ident(slot_name)
+
+        if slot_type == REPLICATION_LOGICAL:
+            command += "LOGICAL "
+        elif slot_type == REPLICATION_PHYSICAL:
+            command += "PHYSICAL "
+        else:
+            raise psycopg2.ProgrammingError("unrecognized replication slot type")
```
ringerc: "unrecognized replication slot type %s" % slot_type

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+
+        if start_lsn is None:
+            start_lsn = '0/0'
+
+        # reparse lsn to catch possible garbage
+        lsn = start_lsn.split('/')
+        command += "%X/%X" % (int(lsn[0], 16), int(lsn[1], 16))
```
ringerc: IMO the LSN should internally  be a uint64, formatted in Pg's x/y format for output on the wire.

lib/extras.py
```diff
@@ -437,6 +438,150 @@ def callproc(self, procname, vars=None):
...
+                if not command.endswith('('):
+                    command += ", "
+                command += "%s %s" % (self.quote_ident(k), _A(str(v)))
+            command += ")"
+
+        return self.start_replication_expert(command, writer=writer,
+                                             keepalive_interval=keepalive_interval)
+
+    def send_feedback_message(self, written_lsn=0, sync_lsn=0, apply_lsn=0, reply_requested=False):
+        return self.send_replication_feedback(written_lsn, sync_lsn, apply_lsn, reply_requested)
```
ringerc: Why are both `send_feedback_message` and `send_replication_feedback` methods defined?

psycopg/cursor.h
```diff
@@ -72,6 +73,14 @@ struct cursorObject {
...
 #define DEFAULT_COPYBUFF  8192
 
+    int         repl_stop;             /* if client requested to stop replication */
+    struct timeval repl_keepalive_interval;   /* interval for keepalive messages in replication mode */
+    XLogRecPtr  repl_write_lsn;        /* LSN stats for replication feedback messages */
+    XLogRecPtr  repl_flush_lsn;
+    XLogRecPtr  repl_apply_lsn;
+    int         repl_feedback_pending; /* flag set when we couldn't send the feedback to the server */
+    struct timeval repl_last_io;       /* timestamp of the last exchange with the server */
+
```
ringerc: This block should have comments indicating that it relates to `ReplicationCursor` and `ReplicationConnection`, if it can't be moved to a subclass.

psycopg/cursor_type.c
```diff
@@ -35,9 +36,11 @@
 #include "psycopg/microprotocols_proto.h"
 
 #include <string.h>
-
```
ringerc: Unnecessary whitespace change should be removed

psycopg/cursor_type.c
```diff
@@ -1579,6 +1582,185 @@ psyco_curs_copy_expert(cursorObject *self, PyObject *args, PyObject *kwargs)
...
+
+#define psyco_curs_stop_replication_doc \
+"stop_replication() -- Set flag to break out of endless loop in start_replication() on sync connection."
+
+static PyObject *
+psyco_curs_stop_replication(cursorObject *self)
+{
+    EXC_IF_CURS_CLOSED(self);
+
+    self->repl_stop = 1;
```
ringerc: Should this do any tests to behave differently in sync vs aync mode?

psycopg/libpq_support.h
```diff
@@ -0,0 +1,51 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+ * License for more details.
+ */
+#ifndef PSYCOPG_LIBPQ_SUPPORT_H
+#define PSYCOPG_LIBPQ_SUPPORT_H 1
+
+#include "psycopg/config.h"
+
+/* type and constant definitions from internal postgres includes */
+typedef unsigned int uint32;
```
ringerc: This isn't portable, or really correct.

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
     return ret;
 }
 
+/* ignores keepalive messages */
```
ringerc: This comment doesn't really sum up what the function does ;)

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+           server we might be reading a number of messages for every single
+           one we process, thus overgrowing the internal buffer until the
+           client system runs out of memory. */
+        if (!PQconsumeInput(curs->conn->pgconn)) {
+            pq_raise(curs->conn, curs, NULL);
+            goto exit;
+        }
+        /* But PQconsumeInput() doesn't tell us if it has actually read
+           anything into the internal buffer and there is no (supported) way
+           to ask libpq about this directly.  The way we check is setting the
```
ringerc: Sounds like it's worth fixing that in libpq down the track.

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+           flag and re-trying PQgetCopyData(): if that returns 0 again,
+           there's no more data available in the buffer, so we return None. */
+        consumed = 1;
+        goto retry;
+    }
+
+    if (len == -2) {
```
ringerc: Comments on these failure cases would be helpful

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+        Py_DECREF(str);
+        if (!msg) { goto exit; }
+
+        ((replicationMessageObject *)msg)->data_start = data_start;
+        ((replicationMessageObject *)msg)->wal_end = wal_end;
+        ((replicationMessageObject *)msg)->send_time = send_time;
+    }
+    else if (buffer[0] == 'k') {
+        /* msgtype(1), walEnd(8), sendTime(8), reply(1) */
+        hdr = 1 + 8 + 8;
```
ringerc: comment here doesn't match the code, and there's no `reply` field in a walsender protocol message

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+        if (len < hdr + 1) {
+            psyco_set_error(OperationalError, curs, "keepalive message header too small");
+            goto exit;
+        }
+
+        reply = buffer[hdr];
```
ringerc: this appears to access the first byte after the walsender protocol header. This isn't defined as meaning any particular thing AFAIK. I don't understand what's being done here.

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+        if (reply) {
+            if (!pq_send_replication_feedback(curs, 0)) {
+                if (curs->conn->async) {
+                    curs->repl_feedback_pending = 1;
+                } else {
+                    pq_raise(curs->conn, curs, NULL);
```
ringerc: This could use an error message indicating the context in which the underlying error occurred. "While sending replication feedback" or similar.

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+        msg = PyObject_CallFunctionObjArgs((PyObject *)&replicationMessageType,
+                                           curs, str, NULL);
+        Py_DECREF(str);
+        if (!msg) { goto exit; }
+
+        ((replicationMessageObject *)msg)->data_start = data_start;
+        ((replicationMessageObject *)msg)->wal_end = wal_end;
+        ((replicationMessageObject *)msg)->send_time = send_time;
+    }
+    else if (buffer[0] == 'k') {
```
ringerc: /* Primary keepalive message */

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+    /* It also makes sense to set this flag here to make us return early in
+       case of retry due to keepalive message.  Any pending data on the socket
+       will trigger read condition in select() in the calling code anyway. */
+    consumed = 1;
+
+    /* ok, we did really read something: update the io timestamp */
+    gettimeofday(&curs->repl_last_io, NULL);
+
+    Dprintf("pq_read_replication_message: msg=%c, len=%d", buffer[0], len);
+    if (buffer[0] == 'w') {
```
ringerc: /* XLogData message, i.e. replication data payload */

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+    if (PQputCopyData(curs->conn->pgconn, replybuf, len) <= 0 ||
+        PQflush(curs->conn->pgconn) != 0) {
+        return 0;
+    }
+    gettimeofday(&curs->repl_last_io, NULL);
+
+    return 1;
+}
+
+/* used for streaming replication only */
```
ringerc: Is this also only used in synchronous mode? The while(1) below suggests so, but it could be made explicit.

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+
+    /* if the file is text we must pass it unicode. */
+    if (-1 == (is_text = psycopg_is_text_file(curs->copyfile))) {
+        goto exit;
+    }
+
+    CLEARPGRES(curs->pgres);
+    pgconn = curs->conn->pgconn;
+
+    while (1) {
```
ringerc: /*
 * This is synchronous mode ReplicationCursor.start_replication(...) loop, which consumes messages
 * and sends feedback and keepalives until asked to stop, feeding them to a user-provided callback
 * object.
 */

psycopg/pqpath.c
```diff
@@ -1514,6 +1529,280 @@ _pq_copy_out_v3(cursorObject *curs)
...
+            fd = PQsocket(pgconn);
+            if (fd < 0) {
+                pq_raise(curs->conn, curs, NULL);
+                goto exit;
+            }
+
+            FD_ZERO(&fds);
+            FD_SET(fd, &fds);
+
+            gettimeofday(&curr_time, NULL);
```
ringerc: `/* Do we need to send feedback? */`

psycopg/libpq_support.h
a1exsh: Well, that's how it is defined in c.h in postgres code.  Actually, it appears to me now, that we can just `#include "internal/c.h"`, as it gets installed as part of libpq-dev on Debian, for example.

lib/extras.py
a1exsh: I don't recall at the moment what stopped me from implementing a subclass of cursor type, but that is the reason.  The one with "replication" in the name is defined on the general cursor type, and the other one is just a thin wrapper/alias on the `ReplicationCursor` type.  Definitely implementing a proper subclass would be better approach.

lib/extras.py
a1exsh: If we would store this in an object attribute or pass the arg to C code, yes.  But it is so much easier to just construct the replication protocol command in Python, so there's no actual difference: we append it to the string and that's it.

ringerc:
Sorry about the comment spam above; was doing line-by-line review/annotation.

Overall this looks really, really good. A few nitpicks here and there, but my only major concern turned out to be more of a docs issue - the API and implementation look just like what I wanted and would've wanted to do.

Excellent to see detailed documentation, comments, and examples.

I was about to start implementing this, but you've beaten me to it and no doubt done it lots better than  I would've.

I'd love to see this merged with the changes noted above.

lib/extras.py
a1exsh: Good point.  These are not text now, but this is just a leftover of C implementation of these functions (you can see them `#defined` in cursor.h but never used).

a1exsh:
Thanks again for your to-the-point comments!  This was sitting around with little attention for a while. :)

lib/extras.py
a1exsh: Slot name is optional in case of `START_REPLICATION`, but not with `CREATE_REPLICATION_SLOT`.

ringerc:
If you get a chance to update the pull request I'd love to do some testing of it against the logical decoding plugin I'm developing.

a1exsh:
I'm on it, expect an updated version before end of this week.

doc/src/extras.rst
PJMODOS: btw non-asynchronous = synchronous connection :)

a1exsh: Yes.  I have no idea why I've started using that term.  It doesn't make docs any better.

doc/src/extras.rst
a1exsh: Well, I'm not so sure about that.  The python's `select` doc says that on Windows it only works with sockets, but this is exactly our case: https://docs.python.org/2/library/select.html

doc/src/extras.rst
a1exsh: Excellent point, thanks for spotting it!

a1exsh:
Hey Craig, I've updated this pull request.  I was trying to address all of the feedback, but could do only this much:

* support for 9.3 by allowing `replication=true` instead of `replication=database` conninfo parameter
* this provoked splitting the connection class in two, otherwise there was no good option to pass the replication type from the outside
* also taken the loop from `start_replication` out to a separate function `consume_replication_stream`, only intended for the sync connections
* documentation updated and fixed accordingly

Didn't have the time to try to make replication cursor a proper subclass yet.

Happy hacking! :-)

ringerc:
Brilliant, thanks very much. I'll give it a try early this week on the output plugin I'm working on. 

(I should be able to give you git access, but I'll have to check - it's PostgreSQL licensed, but won't be released to the public for a month or so yet. Email me directly as `__import__('base64').decodestring("Y3JhaWcucmluZ2VyY0AybmRxdWFkcmFudC5jb20=")`).

Good idea splitting into logical and physical replication connections.

I like the `consume_replication_stream` change. Makes the API easier to understand by avoiding one function with very different behaviour depending on connection options set distantly from the function.

a1exsh:
I'm away from keyboard the whole next week, so I won't be able to check your plugin out. By the way for our purposes we were working on a plugin to produce JSON output, would that be of interest for you?

ringerc:
Yes, I think a plugin for json output would be very useful.

It's different to what I'm doing - what I'm involved with is focused on a compact, efficient binary stream for use in logical replication, feeding message queue systems, etc. It's an outgrowth of the output plugin developed for BDR. A json text representation is unsuitable for this for a lot of reasons.

A json-producing plugin would be really good as a general purpose utility, though, and I'd love to see one included in contrib. I think there are few people working on them though, so try to avoid duplicating efforts.

ringerc:
Before merging I think the connection should preferably be split into a proper subclass, though it might not be worth the verbosity involved when working with the CPython API.

I also suspect that the row-by-row COPY support should be pushed down to be shared with copy_expert. There's so much logic that'd be shared: the push (callback) or pull oriented read modes, the support for reading raw bytes or decoded unicode text for rows, the result object, etc. The replication message class would have a few extra fields that a copy result wouldn't, but that's pretty much the limit of the differences. That said, I don't think it's particularly important to do this now - it should be possible to do it as a later refactor if I or someone else have the time and interest without breaking clients of the proposed replication API.

I'm pulling this to test it out now.

ringerc:
```
psycopg/replication_message_type.c: In function ‘replmsg_repr’:
psycopg/replication_message_type.c:54:9: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 8 has type ‘pg_int64’ [-Wformat=]
         self->send_time);
         ^
```

doc/src/extras.rst
```diff
@@ -141,6 +141,383 @@ Logging cursor
...
+                              used by the slot; required for logical
+                              replication connections, disallowed for physical
+
+        Example::
+
+            log_cur.create_replication_slot("logical1", "test_decoding")
+            phys_cur.create_replication_slot("physical1")
+
+            # either logical or physical replication connection
+            cur.create_replication_slot("slot1", slot_type=REPLICATION_LOGICAL)
```
ringerc: The signature documented for create_replication_slot above lacks slot_type argument, though one is present in the help and in the actual function definition:

```
Help on ReplicationCursor in module psycopg2.extras object:

class ReplicationCursor(psycopg2.extensions.cursor)
 |  A cursor used for communication on the replication protocol.
 |  
...
 |  create_replication_slot(self, slot_name, slot_type=None, output_plugin=None)
 |      Create streaming replication slot.
```

If the `LogicalReplicationConnection` and `PhysicalReplicationConnection` separation exists then this param shouldn't need to exist.

However, arguably that separation shouldn't be `LogicalReplicationConnection` and `PhysicalReplicationConnection` but rather `ReplicationConnection` and `DatabaseReplicationConnection`. Currently `replication=database` isn't used for anything else, but it might well be in future. Then `create_replication_slot` could take a `slot_type` argument, with either value legal for a `DatabaseReplicationConnection` and with only `REPLICATION_PHYSICAL` legal for a base `ReplicationConnection`.

Reasonable? Or am I over-complicating things? It seems like this would be a closer mapping to the underlying PostgreSQL API.

a1exsh: I've first removed the slot_type parameter, but then noticed that one can still create a slot of either type on either type of connection.

Not sure if proposed connection classes names make it any clearer, but it does map closer to the interface, I agree.

psycopg/pqpath.c
```diff
@@ -1575,6 +1876,20 @@ pq_fetch(cursorObject *curs, int no_result)
...
+        Dprintf("pq_fetch: data from a streaming replication slot (no tuples)");
+        curs->rowcount = -1;
+        ex = 0;
+        /*if (curs->conn->async) {
+            ex = 0;
+        } else {
+            ex = _pq_copy_both_v3(curs);
+            
+            if (PyErr_Occurred()) ex = -1;
+        }*/
```
ringerc: Commented out code should be deleted

psycopg/pqpath.c
ringerc: Still think this needs a comment.

doc/src/extras.rst
```diff
@@ -141,6 +141,383 @@ Logging cursor
...
+
+        :param reply: request the server to send back a keepalive message immediately
+
+        This method tries to flush the latest replication feedback message
+        that `send_replication_feedback()` was trying to send but couldn't.
+
+        If *reply* is `!True` sends a keepalive message in either case.
+
+        Returns `!True` if the feedback message was sent successfully,
+        `!False` otherwise.
```
ringerc: `!True`? `!False`? What?

Also, should IMO return `True` if called without `reply` set and there are no queued feedback messages. It successfully sent the zero messages that were queued, that's not a failure.

doc/src/extras.rst
```diff
@@ -141,6 +141,383 @@ Logging cursor
```
ringerc: Quick start for docs:
---

You must be using PostgreSQL 9.4 or above to run this quick start.

Make sure that `replication` connections are permitted for user `postgres` from `localhost` in `pg_hba.conf` then `SELECT pg_reload_conf()`. Create a database `psycopg2test`.

Then run something like the following to quickly try the replication support out. This is not production code - it has no error handling, it sends feedback too often, etc - and it's only intended as a simple demo of logical replication functionality.

```
import psycopg2
import psycopg2.extras

conn = psycopg2.connect('dbname=psycopg2test user=postgres',
   connection_factory=psycopg2.extras.LogicalReplicationConnection)
cur = conn.cursor()
try:
    cur.start_replication(slot_name = 'pytest')
except psycopg2.ProgrammingError:
    cur.create_replication_slot('pytest',
        slot_type=psycopg2.extras.REPLICATION_LOGICAL,
        output_plugin='test_decoding')
    cur.start_replication(slot_name = 'pytest')

class DemoConsumer(object):
    def consume(self, msg):
        print msg.payload
        msg.cursor.send_replication_feedback(flush_lsn=msg.data_start)

democonsumer = DemoConsumer()

cur.consume_replication_stream(democonsumer)
```

This will continue running until cancelled with control-C.

You can now make changes to the `psycopg2test` database using a normal `psycopg2` session, `psql`, etc and see the logical decoding stream printed by this demo client.

doc/src/extras.rst
a1exsh: That's just some weird reStructuredText syntax. Good point about sending zero messages, it would make sense to return success here.

ringerc:
It works well
---

The connection defaults to autocommit on, as is required.

replication=database is added correctly.

Creation and dropping of slots works fine.

Even in synchronous mode I can consume changes using `cur.read_replication_message()` without blocking, which is nice, and was a pleasant surprise.

Sending a keepalive with `send_replication_feedback()` seems reasonable.

Feedback seems to work fine. Forcing a disconnect without feedback causes messages to be replayed again, as expected. If feedback is sent and has been received by the server then the confirmed section of stream is not replayed again. Good.

Consuming the change stream with `consume_replication_stream` in sync mode works well.

Issues:

Async mode broken?
---

or I'm using it wrong, but I'm doing what [the documentation suggests](http://initd.org/psycopg/docs/advanced.html#asynchronous-support):

```
>>> conn = psycopg2.connect('dbname=craig user=postgres', connection_factory=psycopg2.extras.LogicalReplicationConnection, async=1)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/psycopg2/__init__.py", line 164, in connect
    conn = _connect(dsn, connection_factory=connection_factory, async=async)
  File "/usr/lib64/python2.7/site-packages/psycopg2/extras.py", line 496, in __init__
    super(LogicalReplicationConnection, self).__init__(*args, **kwargs)
TypeError: __init__() takes exactly 2 arguments (3 given)
```

also, the async section of the replication docs really should link to that page.


Minor test notes
---

It permits autocommit to be turned off:
===

```
>>> conn.autocommit
True
>>> conn.autocommit = False
>>> cur.drop_replication_slot('test')
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/psycopg2/extras.py", line 545, in drop_replication_slot
    self.execute(command)
psycopg2.ProgrammingError: syntax error
```

but really, that's a "don't do that", not something that needs code changes.

Similarly, `cur.execute` is still allowed, though most SQL commands won't work. This is useful to allow custom replication commands to be run if the psycopg2 interface doesn't know about them, so it should be retained.

Many operations on the cursor class fail with a syntax error
===

e.g.

```
>>> conn.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_READ_UNCOMMITTED)
  File "<stdin>", line 1, in <module>
psycopg2.ProgrammingError: syntax error
```

because replication connections just don't have those attributes. These don't cause problems with correct use of the replication connection, so if someone cares later they can make the method throw NotSupportedException, but I don't think it's worth caring urgently.

To my surprise, `conn.get_backend_pid()` works on a replication connection, which is nice.

The tpc_ methods already fail with `tpc_begin can't be called in autocommit mode` so if !autocommit is disallowed they'll fail as they should.


Replication timeouts aren't graceful
===

The default walsender timeout is quite short, so the server will emit:

```
LOG:  terminating walsender process due to replication timeout
```
and the client will see

```
>>> cur.read_replication_message()
  File "<stdin>", line 1, in <module>
psycopg2.DatabaseError: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

IIRC when this happens PostgreSQL sends a proper connection close message, it doesn't just disconnect the socket. So it should be possible to report "Connection closed by server, probably by administrator action or connection timeout".

This is really another cosmetic point though.

a1exsh:
Oops about async mode - haven't found the time to test that after the changes. :-(  Looks like some minimal unit tests would be useful here.

ringerc:
Agreed on unit tests. I'll take a look at how pyscopg2's tests work shortly.

First I'm going to plug this into the output plugin I'm working on as an alternative to the current SQL-interface tests, which will be useful since it's a binary-mode output plugin.

ringerc:
Minor API niggle: `cur.identify_system()` consumes and returns the result set, but `create_replication_slot` returns no value and its result set must be obtained with `fetchone()` or `fetchall()`.

a1exsh:
Yeah, I had the intent to rectify that. Not sure if the first approach
works with async mode though.

ringerc:
There's a bug in replication state tracking when an attempt to start replication fails. If slot `pytest` does not exist, then this:

```
try:
   cur.start_replication(slot_name = 'pytest')
except psycopg2.ProgrammingError, ex:
   print(ex)
cur.start_replication(slot_name = 'pytest')
```

fails with 

```
replication slot "pytest" does not exist

  File "/tmp/test.py", line 16, in <module>
    cur.start_replication(slot_name = 'pytest')
  File "/usr/lib64/python2.7/site-packages/psycopg2/extras.py", line 597, in start_replication
    return self.start_replication_expert(command)
psycopg2.ProgrammingError: replication already in progress
```

psycopg/cursor_type.c
```diff
@@ -1579,6 +1583,219 @@ psyco_curs_copy_expert(cursorObject *self, PyObject *args, PyObject *kwargs)
...
+
+    gettimeofday(&self->repl_last_io, NULL);
+
+    if (pq_execute(self, command, self->conn->async,
+                   1 /* no_result */, 1 /* no_begin */) >= 0) {
+        res = Py_None;
+        Py_INCREF(res);
+    }
+
+    self->repl_started = 1;
```
ringerc: I think this should be 

```
} else {
    self->repl_started = 1;
}
```

so that `repl_started` is not set if `pq_execute` returns an error.

a1exsh: Correct.

ringerc:
Thanks for your patience with all this. I hope I'm not being too nitpicky, and raising mostly useful points.

I think it's confusing that `start_replication` and `stop_replication` don't really seem to be symmetrical.

`start_replication` issues a `START_REPLICATION ...` command, putting the connection into `COPY BOTH` mode and making it ready to receive replicated tuples.

`stop_replication` does not take the connection out of `COPY BOTH` mode. It just sets a flag that stops the `consume_replication_stream` loop if it's in progress. It doesn't appear to end `COPY BOTH` mode.

So this:

```
try:
   cur.consume_replication_stream(democonsumer)
except KeyboardInterrupt,ex:
   cur.stop_replication()
   cur.drop_replication_slot('pytest')
```

will fail with

```
psycopg2.OperationalError: PQexec not allowed during COPY BOTH
```

I'm checking into how (if?) one can cleanly exit `copy both` mode in `libpq`.

a1exsh:
Yeah, I don't think there was a clean way to do that last time I checked.
This is likely a problem in the server-side also, as it is not prepared to
receive commands after copy both has started.

ringerc:
I suspect a fair bit more is required to make this replication support fully handle the physical replication protocol btw. I'm not bothered, personally, it's perhaps just worth noting in docs that support for `PHYSICAL` mode streaming may be incomplete.

ringerc:
I think you're right re ending streaming - in the code above, `pg_recvxlog` is acknowledging the server ending the stream by sending its own `PQputCopyEnd`. It won't work when the server is still streaming data. So once the connection is streaming, it'll stay streaming.

ringerc:
perhaps `stop_replication` should be `stop_consuming_replication_stream`?

a1exsh:
Well, that's over my personal limit for verbosity for sure ;-)

ringerc:
Me? Verbose? Never.

No strong opinion on the name, other than that it shouldn't be `stop_replication` if it does't reverse what `start_replication` does / put the connection back in normal state.

doc/src/extras.rst
```diff
@@ -141,6 +141,383 @@ Logging cursor
...
+            load on network and the server.  A possible strategy is to confirm
+            after every COMMIT message.
+
+    .. method:: stop_replication()
+
+        This method can be called on synchronous connections from the
+        ``consume()`` method of a ``consumer`` object in order to break out of
+        the endless loop in `consume_replication_stream()`.  If called on
+        asynchronous connection or outside of the consume loop, this method
+        raises an error.
```
ringerc: It doesn't raise an error when called outside the consume loop.

doc/src/extras.rst
```diff
@@ -141,6 +141,383 @@ Logging cursor
...
+                consumer.consume(msg)
+            else:
+                timeout = keepalive_interval - (datetime.now() - cur.replication_io_timestamp).total_seconds()
+                if timeout > 0:
+                    sel = select.select([cur], [], [], timeout)
+                else:
+                    sel = []
+
+                if not sel:
+                    cur.send_replication_feedback()
```
ringerc: On timeout, `select.select()` returns `([], [], [])` which evaluates as `True` since it's a non-empty tuple. I think this should probably be

```
                else:
                    sel = ([],[],[])

                if not sel[0]:
                    cur.send_replication_feedback()
```

a1exsh: Yes, makes sense. Thank you for the comments, they are much appreciated!

ringerc:
I've been able to test the replication support with the decoding plugin I'm working on. It's working perfectly for handling the plugin's binary output over the walsender protocol.

a1exsh:
The problem with the whole approach of doing this:
```py
try:
  cur.start_replication(...)
  ...
except:
  cur.drop_replication_slot(...)
```
is that once you've issued `START_REPLICATION ...`, the server is unprepared for any more commands.  You pretty much need a separate connection to issue `DROP_REPLICATION_SLOT ...` or `SELECT pg_drop_replication_slot(...)`.  This is what I ended up doing in the tests I'm started to write now: https://github.com/zalando/psycopg2/commit/cf4f2411bfd2d5a1cb84393f135e48107428137b#diff-b6a5138ef110c506795c0ac14a4ee9d8R39

Now that I think about this, one should avoid opening a replication connection in order to just drop a slot: the server might be out of `max_wal_senders`, so `pg_drop_replication_slot()` should be preferred.  For the tests, this shouldn't matter much, but in real life this can make a difference.  We should add this to the docs in some form.

ringerc:
Good point.

I've done the same in the tests I'm using in my output plugin, I just
forgot to update this.

a1exsh:
Looking at `WalSndCheckTimeOut` which logs that message and calls `WalSndShutdown` I can see there is intentionally no attempt to notify the client gracefully.  So from the client perspective this is not really different from an actual backend crash.

May be we can change this on the server, so it will try to send at least something.  This doesn't hurt if the client has terminated.

a1exsh:
@ringerc I'm inclined to remove the `stop_replication` method completely, in favor of raising an exception as *the* way to break out of the consumer loop.

It supposed to be a rare occurrence and it happens on uhandled errors anyway.  In testing I expect it to happen with `KeyboardInterrupt` most of the time.

As a side effect we remove the need for a separate `stop_repl` flag, which is good.  Makes sense?

ringerc:
Sounds sensible to me.​

a1exsh:
With the latest commit 0bb81fc, the too verbose `replication_` part is removed from a number of methods on a replication cursor.  Except for methods that map to the replication protocol directly such as `create/drop_replication_slot` and `start_replication`.

a1exsh:
Overall, I think we're pretty close.  I will add more tests and re-read the doc for potential improvements, e.g. incorporate the proposed quickstart section.

a1exsh:
There's a small breakage introduced with commit e69dafb, but I believe it makes more sense that way.

ringerc:
I'll update to this and test it against pglogical_output shortly. That's the output plugin I alluded to; see [the commitfest entry](https://commitfest.postgresql.org/7/418/) for details if you're interested. Since you're doing this work on logical decoding I think it could be useful to you.

a1exsh:
@ringerc cool, thanks.  I think you'll need quite a minor fix (`s/read_replication_message/read_message/` and remove `decode=False` as it's now moved to `start_replication` and defaults to `False` as it should).

While at it, can I also ask you to express your opinion on #363 which suggests basis for moving `ReplicationConnection` object to C level: https://github.com/zalando/psycopg2/commit/fbcf99ad070a3eae67c258d357ab86bda29793fd ?

Thanks!

ringerc:
Hi. I haven't lost track of this, just haven't had a chance to revisit it. Unfortunately the quirks of PostgreSQL core development have forced me to drop work on the Python based tests in favour of "good" ol' pg_regress despite the numerous deficiencies of this approach.

I'm planning a conference presentation where I'd like to make reference to this work at FOSDEM PgDay 2016; feel free to get in touch directly if you're attending or interested in what I'm thinking of doing. You have my email from earlier.

valgog:
Guys, are there any open topics that stop this pull request to be merged?

dvarrazzo:
I haven't followed actively the development of this branch. @a1exsh and @ringerc are the best people to answer your question.

a1exsh:
@dvarrazzo @ringerc @valgog for me there is just one thing I'd like to see, namely: `ReplicationConnection` as a CPython object.  It has been implemented here: https://github.com/zalando/psycopg2/compare/feature/replication-protocol...zalando:feature/replication-protocol-c-connection-object?expand=1, but requires a decision to merge #363 first.

I'm still OK if we don't go that route and just merge this one.

a1exsh:
@dvarrazzo I've pushed the `ReplicationConnection` class based on your version of `make_dsn()`.  I believe this feature is complete now.

Though you might want to squash some commits -- not sure why my implementation also shows up as removed lines, I wanted to avoid rebase a bit too much it seems ;-)

LappleApple:
Hi @dvarrazzo and @fogzot -- colleague of @a1exsh and @valgog here (I manage our open-source culture and projects). Just checking in on @a1exsh's behalf to see when you might merge his pull request here.  Could you let us know, or just go ahead and merge if all's clear? Thanks!

valgog:
@dvarrazzo is there something that needs to be done for this PL be merged?

psycopg/replication_message_type.c
```diff
@@ -0,0 +1,190 @@
...
+    return res;
+}
+
+#define OFFSETOF(x) offsetof(replicationMessageObject, x)
+
+/* object member list */
+
+static struct PyMemberDef replicationMessageObject_members[] = {
+    {"cursor", T_OBJECT, OFFSETOF(cursor), READONLY,
+        "TODO"},
```
dvarrazzo: to do

dvarrazzo:
Hi,

The thread was super-heated and suddenly went quiet after @ringerc dropped his work, and @a1exsh has been silent since November (as far as this thread is concerned). I don't know much about streaming replication so I don't really know what are the desired features set and how this implementation matches them.

What I can do is just check that the documentation matches reality and the tests pass, and that there aren't really glaring idiosyncrasies in the interface. If all the people involved in this thread agree with the current state of the affair I can merge this branch into master and schedule it for 2.7 release.

As for the time scale, I'm sorry but I'm super-swamped with my job and I have hardly found time for the 2.6.2 release, which is frankly more urgent (Py 3.5 support). I can try and see if can do some work across this weekend, but I cannot promise that, and ask for some timeslot at work to work on psycopg but that can only happen after April.

So, as @valgog, @a1exsh, @LappleApple are concerned, are you all happy about the current state of the feature?

psycopg/replication_message_type.c
a1exsh: Whoopsie ;-)

a1exsh:
@dvarrazzo thanks for looking at this again :-)

The interface was stable since last November, yes (if you ignore the fact that there was only a handful of users).  We are trying to promote this work by speaking at conferences, etc., but unfortunately it doesn't seem like it has received a whole lot of attention...

As far as I'm concerned, I'm pretty happy with the current state, especially since we've merged that `make_dsn()` part recently.

Thanks you!

valgog:
And it is me again. @dvarrazzo did you have a chance to make give a final look on that?

Would it be possible to move forward with merging this PR? Anything that stops us from doing it?

PJMODOS:
I've been discussing this patch with @dvarrazzo at pgconf.uk and I just want to repeat my opinion here. I have been using this for a while in testing scripts for pglogical and am generally happy with it in the current form. I think it's ready for merge.

dvarrazzo:
Merged to master. Testing would be appreciated. Thank you very much to all the contributors!

PR #328, opened 2015-06-08
Fix Windows 64bit lobject support for very (>2GB) large objects

jerickso:
The type 'long' with Windows Visual C is 32bits in size for both 32bit and 64bit platforms.  Changed type of variables that could be > 2GB from long to Py_ssize_t.


dvarrazzo:
Merged, thank you very much

PR #341, opened 2015-08-11
Import ujson when available for speed improvements on large hstore

mcclurem:
I recently ran into a situation where my ORM was inappropriately loading large hstore hashes when I wasn't necessarily going to use them. As a result, I started debugging performance of a json load from psycopg2 and found that I could get dramatic speedup when I /did/ want to load these large hstore dictionaries by using ujson instead of the base json module.
Adding an optional dependency of ujson should allow users who want it to improve their hstore and json type performance.
One caveat:
I don't know how you want this handled in terms of the test suites.
Is there a clever way to write the unit tests so that you can run the unit tests twice with and without the ujson module available to import? or do we assume that it's "just another testing configuration" - the equivalent of testing on 2.6 versus 2.7?

### DIFF
--- a/lib/_json.py
+++ b/lib/_json.py
@@ -34,13 +34,16 @@ from psycopg2._psycopg import new_type, new_array_type, register_type
 
 
 # import the best json implementation available
-if sys.version_info[:2] >= (2,6):
-    import json
-else:
-    try:
-        import simplejson as json
-    except ImportError:
-        json = None
+try:
+    import ujson as json
+except ImportError:
+    if sys.version_info[:2] >= (2,6):
+        import json
+    else:
+        try:
+            import simplejson as json
+        except ImportError:
+            json = None
 
 
 # oids from PostgreSQL 9.2


dvarrazzo:
On 11 Aug 2015 4:25 pm, "mcclurem" <notifications@github.com> wrote:
>
> Adding an optional dependency of ujson should allow users who want it to
improve their hstore and json type performance.

Hi Morgan,

I appreciate the tip about speeding up json operations. However I don't
want to add the extra dependency, even if optional. I actually think being
optional it could be a problem: in case the module is missing psycopg would
fall back on the default implementation, probably leading to unexpected
performance problems.

Psycopg already provides hooks to plug in a different json implementation:
see the [Json](
http://initd.org/psycopg/docs/extras.html#psycopg2.extras.Json) `dumps`
parameter and the [register_json](
http://initd.org/psycopg/docs/extras.html#psycopg2.extras.register_json)
`loads` parameter which can be used to customize the functions used.

I won't pull this request, but I'd be grateful if you could write a
documentation patch showing how to use these methods to plug `ujson` in. I
would add a subsection or a box in the [json adaptation section](
http://initd.org/psycopg/docs/extras.html#adapt-json) and a FAQ entry (e.g.
Json adaptation is slow: can I use the `super_fast_json` module instead?)
spending a few words and pointing there.

Thank you very much!

PR #350, opened 2015-09-22
Fixes for MSVC 2015 (Python 3.5)

Photonios:
Recently, Python 3.5 was released, which also changed the default compiler used on Windows to MSVC 2015. Psycopg2 did not yet compile under this version of MSVC.

The fixes are simple:

* `isnan` is a macro that MSVC did not define in older version, the code just assumed it wasn't present in all versions, fixed so that it only re-defines it when it is really not existent
* `round` is a function that MSVC did previously not define, now they do, fixed by checking for a specific version

I've never contributed to this project, so please let me know if you want to see things differently :)

### DIFF
--- a/psycopg/config.h
+++ b/psycopg/config.h
@@ -142,15 +142,19 @@ static int pthread_mutex_init(pthread_mutex_t *mutex, void* fake)
 #endif
 #endif
 
+/* what's this, we have no round function either? */
 #if (defined(__FreeBSD__) && __FreeBSD_version < 503000) \
     || (defined(_WIN32) && !defined(__GNUC__)) \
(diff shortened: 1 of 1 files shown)


fogzot:
Seems fine to me. Thank you for the contribution.

dvarrazzo:
It's worth a @jerickso roundtrip...

jerickso:
Looks fine by me.  Also did a quick merge of the patch with the 2.6 branch w/a windows build to make sure I wasn't missing anything obvious, and it passed the build and tests.

dvarrazzo:
Thank you all, I've merged these patches to the 2.6 maintenance branch too.

Yedfree (approved):

PR #353, opened 2015-09-30
Report libpq error code for errors with no message

### LINKED ISSUE #352: psycopg2 treats PGRES_COPY_BOTH as an error
psycopg2 treats the return code `PGRES_COPY_BOTH` as a `DatabaseError`, per #351:

    DatabaseError: error with no message from the libpq

It should probably instead raise `NotImplementedError`, or at least a `ProgammingError` with error message `psycopg2 does not support COPY_BOTH replication mode`

ringerc:
Instead of:

    DatabaseError: error with no message from the libpq

raise a message like:

    DatabaseError: error code PGRES_COPY_BOTH with no error message from
    libpq

(In this particular case PGRES_COPY_BOTH isn't actually an error, but
psycopg2 doesn't understand it and should treat it that way.)

Fixes #352


dvarrazzo:
Fixed in master, to be released in 2.6.1

Issue #308, opened 2015-04-09
Expose libpq PQescapeIdentifier


dvarrazzo:
Allow dynamic query creation by string interpolation without resorting to adaptation (as discussed in #209), which would break switching to PQexecParams.

a1exsh:
How about adding `psycopg2.extensions.QuotedIdentifier` that would mimic the `QuotedString`, but use `PQescapeIdentifier` internallly if available and fall back to `psycopg_escape_identifier_easy` if not?

Since `PQescapeIdentifier` pretty much requires a connection, we could also add a method on connection class, but not sure if that really fits into the picture.  For one, I find that suggestion to expose `quote_ident/literal/nullable` in https://github.com/psycopg/psycopg2/issues/209#issuecomment-90581039 make a lot of sense.

dvarrazzo:
Proposed ideas (exposed in #358):

Add a "compose" method to the cursor to the object returning a sql statement to pass to execute:

```python
from psycopg2.extensions import Identifier

cur.execute(
    cur.compose(
        "insert into %(table)s %(field)s values (%%s)",
        dict(table=Identifier("tbl"), field=Identifier("fld"))),
    [42])
```

Rough ideas are:

* wrapper objects `Identifier`, `Literal` and maybe `String` and `Bytea` would be exposed (names to review, to avoid clashes). These objects derive from a base class e.g. `Composable`.
* `compose()` knows how to adapt `Identifier` and `Literal`. If a normal python object is passed, it is adapted as per `adapt()` and passed as string, going through the regular adaptation machinery.
* `compose()` takes a dict or a string as argument.
* `Composable` objects expose an (e.g.) `adapt()` method taking a connection or cursor and returning the adapted string. The method can be called manually and is used internally by `cur.compose()`.

All in all we have this pseudostuff:

```python
class Composable:
    def __init__(self, wrapped):
        self.wrapped = wrapped
    def adapt(self, conn_or_cur):
        raise NotImplementedError

class Identifier(Composable):
    def adapt(self, conn_or_cur):
        pgconn = get_it_from(conn_or_cur)
        return libpq.PQescapeIdentifier(pgconn, self.wrapped)

# etc for Literal, String, Bytea

class cursor:
    def compose(sql, args):
        adapted = [] # TODO: handle args dict
        for arg in args:
        if isinstance(arg, Composable):
            adapted.append(arg.adapt())
        else:
           # TODO: prepare() etc
[...]

Issue #209, opened 2014-04-21
Create escape literal method


psycoteam:
Originally submitted by: Alex Shafer

I'm currently attempting to create a high-level query builder specific to my application and I need to manually escape literals so they can be added to the query I'm building.

cpburnz:
:+1: It would be convenient if psycopg2 (or the future v3?) provided a utility method similar to PL/Python's [plpy.quote_ident](http://www.postgresql.org/docs/devel/static/plpython-util.html) to escape identifiers that could be used when building a query.

dvarrazzo:
@cpburnz  Wrapping the libpq function to escape identifiers is a good idea: it can be definitely released in one of the next psycopg2 releases.

Half-jokingly, a few days ago I wrote [a recipe to use it from psycopg](https://gist.github.com/dvarrazzo/b7c8f050bbd39dd2c104). It uses ctypes and exploits the members alignment of the psycopg connection structure, so it's not necessarily very robust... but we can expose it no problem in one of the next psycopg versions (not making an adapter out of it, but allowing to use it for string composition).

dvarrazzo:
Merging it to #308.

PR #358, opened 2015-10-13
Add connection.quote_ident(str)

a1exsh:
Also use PQescapeIdentifier to quote cursor name.

See #308 and #209.

### DIFF
--- a/doc/src/connection.rst
+++ b/doc/src/connection.rst
@@ -582,6 +582,16 @@ The ``connection`` class
 
             .. __: http://www.postgresql.org/docs/current/static/libpq-status.html#LIBPQ-PQTRANSACTIONSTATUS
 
+    .. method:: quote_ident(str)
+    
+        Return quoted identifier according to PostgreSQL quoting rules.
+
+        Requires libpq >= 9.0.
+
+        .. seealso:: libpq docs for `PQescapeIdentifier()`__.
+
+            .. __: http://www.postgresql.org/docs/current/static/libpq-exec.html#LIBPQ-PQESCAPEIDENTIFIER
+
 
     .. index::
         pair: Protocol; Version
--- a/psycopg/connection_type.c
+++ b/psycopg/connection_type.c
@@ -733,6 +733,38 @@ psyco_conn_get_parameter_status(connectionObject *self, PyObject *args)
     return conn_text_from_chars(self, val);
 }
 
+/* quote_ident - Quote identifier */
+
+#define psyco_conn_quote_ident_doc \
+"quote_ident(str) -- Quote identifier according to PostgreSQL quoting rules.\n\n" \
+"Requires libpq >= 9.0, which provides PQescapeIdentifier()."
+
+static PyObject *
+psyco_conn_quote_ident(connectionObject *self, PyObject *args)
+{
+#if PG_VERSION_NUM >= 90000
+    const char *str = NULL;
+    char *quoted;
+    PyObject *result;
+
+    EXC_IF_CONN_CLOSED(self);
+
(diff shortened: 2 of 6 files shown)


dvarrazzo:
I really, really, don't like this method to be part of the connection. I think it should be in `extension`, taking the connection as input. The cursor would also be a better place to have it. Another option would be an object internally wrapping the connection and exposing the various escape functions provided by the libpq (string, literal, identifier, bytea), and maybe a composition function similar to the postgres [format](http://www.postgresql.org/docs/9.4/static/functions-string.html#FUNCTIONS-STRING-FORMAT) function (but please, let's talk about that before you rush implementing it).

Another observation: falling back to escape_easy should *not* be supported. If the libpq doesn't offer the feature we don't want to mock it. Conversely, the escape_easy could be reimplemented to use PQescapeIdentifier so that we don't break named cursors for old clients. The function was added to PG 9.0, which is already out of support: I don't really care about people wanting to develop on new features using libpq 8.4.

a1exsh:
Hm... OK.  Do you think it would then make sense not to remove the enclosing quotes in escape_easy and change every place where `curs->name` is used to with `"%s"` format?

Moving `quote_ident` to cursor kinda makes sense.  What about `QuotedIdentifier` I was proposing in #308 -- is this approach OK to you?

dvarrazzo:
Do you want to remove the quotes in order to expose it to Python? In this case no: I'm ok to expose PQescapeIdentifier, not to expose our version.

The QuotedString thingy still requires passing a connection. Using it during adaptation has been proposed several times, but has always been rejected (with reasons) by @fogzot and others.

An approach I'm thinking is to add a "compose" method to the cursor to the object returning a sql statement to pass to execute:

```python
from psycopg2.extensions import Identifier

cur.execute(
    cur.compose(
        "insert into %(table)s %(field)s values (%%s)",
        dict(table=Identifier("tbl"), field=Identifier("fld"))),
    [42])
```

Rough ideas are:

* wrapper objects `Identifier`, `Literal` and maybe `String` and `Bytea` would be exposed (names to review, to avoid clashes). These objects derive from a base class e.g. `Composable`.
* `compose()` knows how to adapt `Identifier` and `Literal`. If a normal python object is passed, it is adapted as per `adapt()` and passed as string, going through the regular adaptation machinery.
* `compose()` takes a dict or a string as argument.
* `Composable` objects expose an (e.g.) `adapt()` method taking a connection or cursor and returning the adapted string. The method can be called manually and is used internally by `cur.compose()`.

All in all we have this pseudostuff:

```python
class Composable:
    def __init__(self, wrapped):
        self.wrapped = wrapped
    def adapt(self, conn_or_cur):
        raise NotImplementedError

class Identifier(Composable):
    def adapt(self, conn_or_cur):
        pgconn = get_it_from(conn_or_cur)
        return libpq.PQescapeIdentifier(pgconn, self.wrapped)

# etc for Literal, String, Bytea

class cursor:
    def compose(sql, args):
        adapted = [] # TODO: handle args dict
        for arg in args:
        if isinstance(arg, Composable):
            adapted.append(arg.adapt())
        else:
           # TODO: prepare() etc
           adapted.append(psycopg2.extensions.adapt(arg).getquoted())

        return sql % adapted

A prototype of that can be created in Python using [my terrible hack](https://gist.github.com/dvarrazzo/b7c8f050bbd39dd2c104) to access the libpq and a mixin subclass of cursor exposing `compose()`, so we can write a test suite and check if the interface makes sense at all.

dvarrazzo:
I'm closing this pull request as what I'd like to see in psycopg is way different from this.

PR #359, opened 2015-10-14
Add psycopg2.extensions.quote_ident.

a1exsh:
@dvarrazzo how about this way? #308

The compose stuff looks nice, but I only need something really simple and straightforward for #322: https://github.com/psycopg/psycopg2/pull/322/files#diff-e6256dfa32f9bced2faae91403f4d594R487

### DIFF
--- a/doc/src/extensions.rst
+++ b/doc/src/extensions.rst
@@ -221,6 +221,19 @@ functionalities defined by the |DBAPI|_.
 
         .. __: http://www.postgresql.org/docs/current/static/libpq-misc.html#LIBPQ-PQLIBVERSION
 
+.. function:: quote_ident(str, scope)
+
+    Return quoted identifier according to PostgreSQL quoting rules.
+
+    The *scope* must be a `connection` or a `cursor`, the underlying
+    connection encoding is used for any necessary character conversion.
+
+    Requires libpq >= 9.0.
+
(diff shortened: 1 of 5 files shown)


dvarrazzo:
I'm fine with it, I can merge it, but I would like to see unicode and Py3 tests.

What does the function return? str or unicode? what on py3? I think it should take both input, encode unicode in the connection encoding (use `psycopg_ensure_bytes()`) and I think it should always emit bytes, but py3 guys may not agree, they like useless encoding/decoding roundtrips, so probably `conn_text_from_chars()` is the right choice, yes, but I would like to see it checked in the test suite.

More notes on the way...

dvarrazzo:
I'm happy to merge this changeset to help you with your work on the streaming replication branch. Of course if we end up improving the quote_ident support as discussed in #308 we can refactor streaming replication to use the new method.

a1exsh:
Thank you!  Will update to address the comments.

a1exsh:
Please see https://github.com/zalando/psycopg2/commit/89bb6b0711f8ba59f7b8e81339ddaa53356233a2

dvarrazzo:
Thank you, that was exactly what I expected.

Good luck with the replication stuff :)

a1exsh:
Cool, many thanks! :)

PR #362, opened 2015-10-27
Fix stale Dprintfs in pqpath.c referring to 'status'

a1exsh:


dvarrazzo:
Huh? How did it even compile? Fixed, thank you.

a1exsh:
Well, you never know if you don't define PSYCOPG_DEBUG :-)

PR #364, opened 2015-10-30
Add connection.get_dsn_parameters()

a1exsh:


fogzot:
Seems like a useful refactor. @dvarrazzo any reason to not merge this?

dvarrazzo:
The patch itself is ok: it merges with the other features @a1exsh has being implementing (e.g. parsing the dsn in C, I seem to remember: bug #363) and haven't really had time to get through all of them. I think we should, to acknowledge his work.

a1exsh:
@dvarrazzo that doesn't work with a closed connection though: the memory gets deallocated by `PQfinish()` which we call from `conn_close[_locked]()`.

dvarrazzo:
Fine, in this case the guard is ok

dvarrazzo:
Isn't this just equivalent to `parse_dsn(conn.dsn)`?

a1exsh:
@dvarrazzo it is almost equivalent, but it will also reveal any *implicit* connection parameters such as `client_encoding` and any parameters which were passed in form of `PG*` environment variables.

For example, you might have completely empty dsn in the connection object, while still overriding the defaults with `PGHOST/PORT/USER/etc`.  This you cannot learn from examining the connection object's `dsn` property, only by querying `PGconn`.

PR #367, opened 2015-11-01
A convenient context manager for connection pools

diabolo-dan:
Based on the ideas presented in: https://github.com/psycopg/psycopg2/pull/17/files
A context manager for connection pools seems like a really useful feature.  My proposal is a convenience method on connection pool that will handle all the levels of context down to the cursor. i.e., usage would be:

<pre>
with my_pool.quick_cursor() as cur:
    cur.execute(prepared_query)
    results = cur.fetchall()
</pre>

This is roughly equivalent to:

<pre>
try:
    connection = my_pool.getconn()
    with connection as conn, conn.cursor as cur:
        cur.execute(prepared_query)
        results = cur.fetchall()
finally:
    my_pool.putconn(connection)
</pre>

Even if the pool was built into it's own context, the three depth of context is annoying, and this seems like a common enough use case that it's worth having an easy and simple way to do it.

This is my preferred method for improving the connection pool situation vis-a-vis context managers.  Combining the pool and connection context, but not the cursor might be preferable for adaptability, but I think for a lot of people this feature is a straightforward way to easily get all the functionality they want.

### DIFF
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -123,7 +123,6 @@ class AbstractConnectionPool(object):
             del self._used[key]
             del self._rused[id(conn)]
 
-    @property
     @contextmanager
     def quick_cursor(self):
         """A ContextManager for quickly getting a cursor"""
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -26,6 +26,7 @@ This module implements thread-safe (and not) connection pools.
 
 import psycopg2
 import psycopg2.extensions as _ext
+from contextlib import contextmanager
 
 
 class PoolError(psycopg2.Error):
@@ -121,6 +122,14 @@ class AbstractConnectionPool(object):
         if not self.closed or key in self._used:
             del self._used[key]
             del self._rused[id(conn)]
+
+    @property
+    @contextmanager
+    def quick_cursor(self):
+        """A ContextManager for quickly getting a  cursor"""
+        with self.getconn() as conn, conn.cursor() as cur:
+                    yield cur
+        self.putconn(conn)
 
     def _closeall(self):
         """Close all connections.
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -126,10 +126,12 @@ class AbstractConnectionPool(object):
     @property
     @contextmanager
     def quick_cursor(self):
-        """A ContextManager for quickly getting a  cursor"""
-        with self.getconn() as conn, conn.cursor() as cur:
-                    yield cur
-        self.putconn(conn)
+        """A ContextManager for quickly getting a cursor"""
+        try:
+           with self.getconn() as conn, conn.cursor() as cur:
+              yield cur
+        finally:
+           self.putconn(conn)
 
     def _closeall(self):
         """Close all connections.


barrust:
When trying to implement this contextmanager, I noticed that the putconn() method was not guaranteed to be called if an exception was raised. I think something like the below could resolve the un-released connection back to the pool.

```python
@property
@contextmanager
def quick_cursor():
    """A ContextManager for quickly getting a  cursor"""
    try:
        conn = self.getconn()
        with conn: # ensure commit or rollback
            with conn.cursor() as cur:
               yield cur
    except:
        raise
    finally:
        self.putconn(conn)
```

diabolo-dan:
Oops, yes, thanks for pointing that out. An obvious oversight on my part.  When the issue came up I made a class to deal with it, and then submitted this seperately, from memory.  I should have tested it, but I wasn't sure if there was enough interest.  (I guess probably not given the responses).

jbagot:
It's a good idea and I'm using but I've added a key parameter because sometimes it's useful put a key in connection.
`@contextmanager
    def get_cursor(self, key=None):
        try:
            with self.getconn(key) as conn, conn.cursor() as cur:
                yield cur
        finally:
            self.putconn(conn, key)`

fake-name:
Any chance this (or something like this) could get merged? 

I was quite surprised that the `psycopg2.pool.ThreadedConnectionPool` *doesn't* support a context-manager interface. That kind of resource management seems like a natural fit to context managers.

dvarrazzo:
@fake-name You are right, it is a good fit, but there is a fundamental mismatch between the meaning of the context managed by the pool and the one managed by the connection, which was analysed in #17. This MR sidesteps the problem by returning a cursor, but this hides elements of controls available on the connection object which IMO shouldn't be hidden (first of all, the transaction management).

So I see this MR more as a recipe, and if your use pattern is compatible with this context manager (i.e. into a `with` block you will run a single atomic set of operation to be committed at the block exit and you won't do anything anymore with that connection) then it's a few line Python function that you can write in your program and use, you can tweak it for small changes etc. Providing it out-of-the-box in the driver suggests an use pattern that I don't think is generic enough.

fake-name:
Having the context-manager return a connection object, rather then a cursor would also be completely fine.

dvarrazzo:
@fake-name it would be better, but it leads to an ugly 3-levels pattern of usage, see #17

calimeroteknik:
>but it leads to an ugly 3-levels pattern of usage, see #17

Isn't that already better than leaking resources?  
That is, forgetting to put the connection back into the pool after use.

Changaco:
I think it's worth mentioning here that [postgres.py](https://github.com/chadwhitacre/postgres.py) provides an abstraction over connection pools. It looks like this:

```python
from postgres import Postgres
db = Postgres('dbname=foo', minconn=1, maxconn=10)

# When you want to commit automatically at the end, or roll back if there's an exception,
# get a cursor:
with db.get_cursor() as cursor:
    rows = cursor.all("SELECT ...")
    row = cursor.one("SELECT ...")
    cursor.run("UPDATE ...")

# When you want full control over the transaction, get a connection instead:
with db.get_connection() as conn:
    cursor = conn.cursor()
    ...
```

(postgres.py isn't a new thing, it actually predates this PR.)

marmelo:
It took me a while to find out why is the connection commited when I return it to the pool, so I decided to share.

After spending some minutes browsing the code I just found a reference in the docs: https://www.psycopg.org/docs/connection.html#connection

> **commit()**
> Commit any pending transaction to the database.
> 
> By default, Psycopg opens a transaction before executing the first command: if commit() is not called, the effect of any data manipulation will be lost.
> 
> The connection can be also set in “autocommit” mode: no transaction is automatically open, commands have immediate effect. See Transactions control for details.
> 
> Changed in version 2.5: **if the connection is used in a with statement, the method is automatically called if no exception is raised in the with block.**

> **rollback()**
> Roll back to the start of any pending transaction. Closing a connection without committing the changes first will cause an implicit rollback to be performed.
> 
> Changed in version 2.5: **if the connection is used in a with statement, the method is automatically called if an exception is raised in the with block.**

I am still looking for the source, please share if you find it.

Thanks.

pauldraper:
> it would be better, but it leads to an ugly 3-levels pattern of usage

I'd prefer 3-levels of usage over the current ugliness.

dvarrazzo:
The psycopg2 pool probably will not receive further improvements. The [psycopg3 pool](https://www.psycopg.org/psycopg3/docs/api/pool.html) offers the interface you would expect.

PR #370, opened 2015-11-06
Modify setup.py to support setuptools/wheel

jerickso:
To support creation of whl files for PyPI, setuptools need to be imported
instead of distutils.  Created try/except case to fall back to integrated
distutils if setuptools is not installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,10 @@ import os
 import sys
 import re
 import subprocess
-from distutils.core import setup, Extension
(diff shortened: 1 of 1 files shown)


heni:
+1 I also interested in this functionality

dvarrazzo:
Merged to my repos to check the full test grid

dvarrazzo:
Merged in 2.6 branch

PR #375, opened 2015-11-10
Update psycopg1.py

kwotuveang3k4bk:
Latest this is working for me - Fedora 23 - python3.4


fogzot:
There is still code depending on psycopg1 API?!
Anyway, this change is absolutely safe: merging it.

PR #383, opened 2015-12-19
Work around late initialization in distutils._msvccompiler.

### LINKED ISSUE #380: Build fails on Windows with Python 3.5, VS 2015
> py -3 setup.py build_ext
    
    [...]
    File "C:\Program Files\Python35\lib\site-packages\setuptools\command\build_ext.py", line 49, in run
      _build_ext.run(self)
    File "C:\Program Files\Python35\lib\distutils\command\build_ext.py", line 338, in run
      self.build_extensions()
    File "C:\Program Files\Python35\lib\distutils\command\build_ext.py", line 447, in build_extensions
      self._build_extensions_serial()
    File "C:\Program Files\Python35\lib\distutils\command\build_ext.py", line 472, in _build_extensions_serial
      self.build_extension(ext)
    File "setup.py", line 308, in build_extension
      '-outputresource:%s;2' % ext_path])
    File "C:\Program Files\Python35\lib\distutils\_msvccompiler.py", line 513, in spawn
      os.environ['path'] = self._paths
    AttributeError: 'MSVCCompiler' object has no attribute '_paths'

I managed to fix this by calling `self.initialize()` at the top of spawn() (see last item in the traceback):

     def spawn(self, cmd):
    +    if not self.initialized:
    +        self.initialize()
         old_path = os.getenv('path')

It builds correctly now, but I suspect the bug may be in psycopg2 rather than distutils. Perhaps you are not supposed to call spawn() on an uninitialized instance?

chrullrich:
Fixes #380.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -301,6 +301,8 @@ class psycopg_build_ext(build_ext):
             except AttributeError:
                 ext_path = os.path.join(self.build_lib,
                         'psycopg2', '_psycopg.pyd')
+            # Make sure spawn() will work if compile() was never
+            # called. https://github.com/psycopg/psycopg2/issues/380
             if not self.compiler.initialized:
                 self.compiler.initialize()
             self.compiler.spawn(
--- a/setup.py
+++ b/setup.py
@@ -301,6 +301,8 @@ class psycopg_build_ext(build_ext):
             except AttributeError:
                 ext_path = os.path.join(self.build_lib,
                         'psycopg2', '_psycopg.pyd')
+            if not self.compiler.initialized:
+                self.compiler.initialize()
             self.compiler.spawn(
                 ['mt.exe', '-nologo', '-manifest',
                  os.path.join('psycopg', manifest),


dvarrazzo:
Could you please add a comment too about why you have to do that manual initialization? (e.g. referencing the bug?)

chrullrich:
Here you go.

dvarrazzo:
I don't think I see the comment... have you pushed it?

chrullrich:
capture

dvarrazzo:
I asked you to add a comment on these statements: commit e80bdc1 has no comment. If you don't do it I'll do myself, I wanted to leave you the attribution.

chrullrich:
Sorry, I misunderstood your request. Done, but who cares about a comment that is not included in the actual repo, but only on the GitHub website?

dvarrazzo:
No, actually I want a comment in `setup.py` :D :D :D

chrullrich:
Sorry, two commits now. Not a Git ace here, either. Can you use this or do you want a single commit?

dvarrazzo:
2 commits is no problem at all, I would use them as they are - that's why you are in a branch. I may only OCD-ly have issues with the commit message of your second comment: I would change into "added comment about msvc initialization" instead (you can change it with git commit --amend and then push --force).

Issue #343, opened 2015-08-13
Deprecate Inet adapter


dvarrazzo:
See issues #205, #342. What is that for, really?

trentclainor:
On SQLAlchemy describe model
```python
from psycopg2.extras import register_cidr
register_cidr()

class Asn(Base):
    __tablename__ = 'asns'
    name = Column(String, nullable=False, unique=True)
    networks = Column(ARRAY(CIDR))
```
Need set cidr[] type in data
```python
from psycopg2.extras import Cidr

networks = ['1.0.4.0/23', '1.0.6.0/24']
asn = Asn(id=asn_id, name=row['name'], data=data, networks=map(Cidr, networks))
```
And later lookup network for ipaddress for found ASN
```python
from sqlalchemy.sql import operators
asn = db.session.query(Asn).filter(Asn.networks.any('1.0.0.1', operator=operators.custom_op('<<'))).first()
```

dvarrazzo:
I assume this was for issue #387.

Anyway, doesn't this work as well using Python strings?

dvarrazzo:
If any, it seems you may want CIDR[] to be recognized as an array (of strings, that would be enough). The following should cast CIDR arrays into list of strings:

```python
psycopg2.extensions.register_type(
    psycopg2.extensions.new_array_type(
        (651,), 'CIDR[]', psycopg2.STRING))
```

would that be already useful to make your loop work?

We may think about having psycopg registering by default the array of known builtin types into array of strings, if there isn't a better adapter for them.

PR #387, opened 2016-01-04
add CIDR and CIDRARRAY types

trentclainor:
Add support cidr[] type in PostgreSQL


dvarrazzo:
We are actually going to drop the Inet adapter too, see #343... Is there a reason for this to exist?

Issue #394, opened 2016-01-19
Backslash escaping changes after creating a connection


JimJty:
I'm generating my own sql for a large data import. I simply wanted to escape some of the strings that I am sending to a template. The first call to the escape function works correctly but subsequent calls return a different string. I'm not sure if I'm using the wrong function or if I ran into a bug. Here is code to reproduce:
```python
    import psycopg2
    import json
    data = {'var1':'20"'}
    str1 = psycopg2.extensions.adapt(json.dumps(data))
    print str1 #returns '{"var1": "20\\""}'  #which is what I want

    #when I make a connection, the results change
    conn = psycopg2.connect(host='localhost', dbname='mydb', user='user1', password='')

    str1 = psycopg2.extensions.adapt(json.dumps(data))
    print str1 #returns '{"var1": "20\""}'  #I lost the escaping of the backslash.    
```
As a workaround, I just wrote my own escape function (which I don't trust at the moment).

dvarrazzo:
Funny.

reduced test case, not requiring json:

```python
import psycopg2

s = '\\'
psycopg2.extensions.adapt(s).getquoted()
"'\\\\'"

conn = psycopg2.connect("")
psycopg2.extensions.adapt(s).getquoted()
"'\\'"
```

fogzot:
I think this is a libpq problem: something is not initialized until one opens the first connection, then everything goes as expected. See #400 for example code.

dvarrazzo:
Yes, I was leaning to thinking that: there is nothing in psycopg that get init'd at the first connection creation. However for sake of sanity I tend to thing the right escaping is the second: I'm not sure why the OP is looking for the first.

fogzot:
I guess the first of is the C-style encoding one can use with E'' strings. After the first connection libpq switches to SQL strings and we don't care because as long as the literals are correctly quoted the backend is happy.

fogzot:
But the problem is: why what seems to be a correctly quoted string is a problem? Doesn't round trip?

JimJty:
As the OP, I wanted the first case because the second case does not produce valid sql. For example, if I am using a template like this:
update core_item set jsonb_col = E{{my_data}}

First case produces (valid):
```sql
update core_item set jsonb_col = E'{"var1": "20\\""}'
```
Second case produces (invalid)
```sql
update core_item set jsonb_col = E'{"var1": "20\""}'
```

dvarrazzo:
[...]

PR #400, opened 2016-01-31
Libq crazyness

fogzot:
Show that #394 is caused by libpq, not psycopg.


dvarrazzo:
Closing the pull request as the `--rpath` thing is not needed with manylinux

PR #409, opened 2016-02-15
Add Python 3 to supported Pythons list (documentation)

kennethreitz:
In my experience, psycopg2 operates quite well on Python 3.5. I'm assuming this is just an oversight, but if there are reasons that it's intentional, sorry for the noise :)


dvarrazzo:
I haven't integrated yet the test suite with Py 3.5: we will declare Py 3.5 officially supported when I will run the test. As the comment one line above says, there are other places where the supported python versions should be advertised.

PR #415, opened 2016-03-09
dont claim copyright for future years

bmwiedemann:
otherwise, when building from unchanged source in 2018,
it would claim Copyright 2018
which is not true

Being able to reproduce identical output from identical input
is important to Linux distributions


dvarrazzo:
mmmmkay.

PR #424, opened 2016-04-18
Avoid a possible null deref, tz might be NULL.

nevun:
Found by clang static analyzer


dvarrazzo:
Cool, thank you.

I've run the static analyzer in the past, but since then a lot of things changed and couldn't run it again when I tried. Have you got a script/recipe to get it running again on psycopg codebase?

nevun:
I did not have to change anything, "scan-build make" just worked out of the box with the latest static analyzer (on a mac).

dvarrazzo:
Thank you, great: I'll try to introduce back a static scan in the tests.

Issue #344, opened 2015-08-13
Deprecate Python 2.5


dvarrazzo:
Dropping its support would make the Py2/Py3 portability easier (no more `b()` "macro" and such).

dvarrazzo:
Py 2.5 dropped in master

PR #429, opened 2016-05-20
syntax update

akstrfn:


dvarrazzo:
I don't think this change is compatible with Python 2.5

The idea is to drop python 2.5 support in 2.7 (see issue #344): in which case this is one of the many cleanup operation we'll have to perform.

Closing now as the issue #344 now has a reference to this and we can remember to do this too (and eventually the other occurrences of deprecated syntax).

PR #445, opened 2016-06-28
Explicitly link with libm

dsilakov:
Without this, psycopg2 fails to compile on systems where "-lm" is not added by default.

Here is one of such examples, though discussion is in Russian:
http://forum.rosalab.ru/viewtopic.php?f=53&t=7272


fogzot:
You can't just add a library like that. What if it doesn't exists? If your build needs any extra libraries you should add them to `setup.cfg`: it exists exactly for that.

dsilakov:
Ok. And what should we advise to users who want to install psycopg2 through pip?

PR #457, opened 2016-07-15
add a __str__ method to NoneAdapter

f0rk:
I'm compiling SQLAlchemy queries and found that when I interpolate the bind parameters, NoneAdapter shows up as <psycopg2.extensions.NoneAdapter object at 0xad8f110> instead of the expected NULL.

### DIFF
--- a/lib/extensions.py
+++ b/lib/extensions.py
@@ -154,6 +154,9 @@ class NoneAdapter(object):
     def getquoted(self, _null=b("NULL")):
         return _null
 
+    def __str__(self):
+        return str(self.getquoted())
+
 
 def make_dsn(dsn=None, **kwargs):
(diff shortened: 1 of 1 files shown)


fogzot:
Can you provide an example of the code that doesn't work? I don't understand the necessity of adding `__str__`.

dvarrazzo:
Can you please raise this bug to SQLAlchemy?

I don't think we should merge this: I think it shows some bad behaviour in SQLAlchemy, and if we condone this the same bug may appear for other types. I think it is better for both the projects to understand why it ends up in that code path.

If we get back from SQLAlchemy and there is a compelling reason for what you propose we could merge this instead (and/or understand if there is something more generic that we should do).

PR #459, opened 2016-07-17
Throw an exception when a NUL character is used as a parameter.

flupzor:


dvarrazzo:
It seems great, thank you

psycopg/utils.c
```diff
@@ -50,8 +50,13 @@ psycopg_escape_string(connectionObject *conn, const char *from, Py_ssize_t len,
     Py_ssize_t ql;
     int eq = (conn && (conn->equote)) ? 1 : 0;
 
-    if (len == 0)
+    if (len == 0) {
         len = strlen(from);
+    } else if (strcspn(from, "\0") != len) {
```
dvarrazzo: I suppose strchr(from, '\0') would be more appropriate as we are looking for a single char

flupzor: I think strcspn is appropriate. strcspn(s, reject) spans until one of the characters in reject is found, not until all of the characters are found. i.e. it is not doing string comparison, it is meant for the case where i'm using it for. I did doubt if I shouldn't use strlen(from), but I decided not to, because with strcspn it is more obvious that we're looking for a NUL terminator inside a string.

strchr returns a pointer, which can also be NULL. So it would be more verbose to implement it like that. i.e. it would read something like, which in my opinion is less readable.
``
char *p = strchr(from, '\0');

if (p != NULL && p - from != len) {
 // exception
} 
``

dvarrazzo: Actually `stren(from) == len` would be a nice trick, but sure, would deserve a line of comment. :D

Actually^2, I don't even get how the `strcspn` implementation can even work using `reject = "\0"`: `strlen(reject)` is 0, not 1, isn't it? I'm not that sure that this use is portable, for some reasonable implementation of `strcspn`.

Any `strcspn` implementation will be more complicated that `strchr` and here we are in a busy loop of the adapter, so I actually prefer to use the more conceptually simple `strchr` and a more "complex" final test. Note however that `man strchr(3)` says that 

> The strchr() and strrchr() functions return a pointer to the matched character or NULL if the character is  not  found.
> The terminating null byte is considered part of the string, so that if c is specified as '\0', these functions return a
> pointer to the terminator.

so ISTM `strchr` will never return NULL using `'\0'` as input char and we only need `if (strchr(from, '\0') != len) { // exception` as a test.

flupzor: Derp, how ironic that I missed that in a fix for a nul-terminator bug. I've adapted the PR to use strchr instead.

dvarrazzo: Thank you, much appreciated!

mattrobenolt:
Hey @flupzor @dvarrazzo, we have just upgraded ourselves to psycopg2 and started getting this exception ourselves in production from user input being blasted at us.

While I understand the behavior that Postgres ends up truncating the data at the null, my concern is centered around how we gracefully handle this in our application now.

We have four real options at the moment:

* Let it error and stop saving the data (least ideal)
* Loop over every parameter ourselves and sanitize for null bytes (pretty heavy handed approach for a small minority of bad data)
* Manually fix every field that we see these as they pop up (probably best, but annoying and sorta whack a mole)
* `try...except ValueError`, then try to sanitize, and recall `execute()` (ALMOST perfect)

We're probably leaning towards option 3 for now since it's not a lot of data, but I'd like to use option 4. So let's address that:

I'd prefer not catching a whole ValueError exception since I'm not entirely sure what else may be triggering a ValueError that would get masked by this behavior. None that I know of, but this seems like it'd be very useful to have a hyper specific exception, such as `NullByteValueError(ValueError)` that we could instead catch, and then proceed to sanitize our values only when we hit this exceptional behavior.

dvarrazzo:
I see from your fix that you are already using a wrapper around the cursor, so you already fixing the issue in a single place, not really whack-a-moling it.

I think it is perfectly fine to catch `ValueError` in `execute()`, trying to sanitise input, and then re-execute, letting it to raise in case it does it again. I don't think we will introduce a specific `ValueError` subclass, but I don't think it's worth worrying because any `ValueError` I can think of would happen before executing the command, not after, so nothing should have side effects and you can re-execute again. Conversely I don't think matching the message is a good idea.

What I think, furthermore, is that you are using a `text` field where you should have used a `bytea` field.

mattrobenolt:
Hey @dvarrazzo, thanks for responding. :)

> I see from your fix that you are already using a wrapper around the cursor, so you already fixing the issue in a single place, not really whack-a-moling it.

This is a Django application, so the cursor is already a bit abstracted away, and that cursor isn't exactly a psycopg2 cursor directly. I ended up doing a slight variation of option 4, where we catch ValueError, then try to fix data before re-submitting. The downside to what we're doing is we opted to be very brittle by checking the exact `e.message` and wrote a test against this behavior so if the message changes in the future, we'd know. Ideally we'd just catch a `NullByteValueError` instead and avoid the brittle-ness of checking the message.

> What I think, furthermore, is that you are using a text field where you should have used a bytea field.

I wish it were that simple. :) [We](https://sentry.io) accept lots of arbitrary data as a part of data ingestions. Fields like "message" and "title" were never intended to be anything other than `text`, but turns out, people send lots of weird data in stacktraces. So here we are. :) I don't disagree that we should do something else to sanitize against this, but we literally didn't even know this was happening until we upgraded, and it's much harder to fix every point of input now, especially since we don't really know the best thing to do with it. (reject, replace chars, etc) considering it was mostly working for our users, I guess as they expected. Our option here to replace `\x00` and escape it, if anything, yields better results in the end since now the string isn't being truncated silently.

> Also, in your patch getsentry/sentry#6960: can't param be a dict?

I don't believe so since this is abstracted through Django, and we haven't seen this in practice yet in production. I suppose you could probably do it if you grabbed the raw cursor an executed by passing a dict, but we apparently don't do this anywhere.

I definitely appreciate your input! <3

Either way, we worked around the issue, was just bringing it up that this was a breaking change for us and required us to patch around it. A ValueError subclass is still what I'd prefer and seems like a cleaner API so I know what error I'm catching, rather than any and all ValueErrors.

PR #482, opened 2016-10-21
do not "SET datestyle" on replication connections

moench-tegeder:
A replication connection - marked by the use of the keyword "replication"
in the DSN - does not support SET commands. Trying to sent "SET datestyle"
will result in an exception.

This was discovered when working against a server with setting "datestyle = german"
with [https://github.com/2ndquadrant-it/barman](url)


dvarrazzo:
Merged, with some improvements and test. Thank you very much for the report.

PR #545, opened 2017-04-16
Fix name of lobject_factory keyword argument

RazerM:


dvarrazzo:
Thank you very much for these contributions :)

Issue #563, opened 2017-06-07
Pool doesn't keep connections open


ibrewster:
When using psycopg2 pool, specifically with the "close" parameter of `putconn` set to false, I was expecting behavior like the following:

- create a "pool" of connections, with an initial number of connections equal to the "minconn" argument
- When getconn is called, see if there is an available connection. If so, return it. If not, open a new connection and return that (up to "maxconn" total connections)
- When putconn is called, return the connection to the pool for re-use, but do *not* close it (unless the close argument is specified as True, documentation says default is False)
- On the next request to getconn, this connection is now available and so no new connection will be made
- perhaps (or perhaps not), after some time, unused connections would be closed and purged from the pool to prevent large numbers of only used once connections from laying around.

However, in my testing, this does not appear to be the case - rather, every time I call `putconn`, the connection is closed unless that would result in fewer than `minconn` connections. Is this expected behavior, or is my expected behavior correct? If this *is* expected behavior, then what is the point of using pool? Just providing a hard limit for the number of allowed simultaneous connections?

dvarrazzo:
Yes, this is the expected behaviour.

The aim of the pool? To allow _minconn_ connection to be ready, it seems. If this is the best policy I have no idea. If there is a better policy feel free to discuss about that in the Mailing List.

PR #565, opened 2017-06-08
Add CachingConnectionPool class

ibrewster:
With existing pool classes, the pool opens a fixed number of connections upon instantiation. If more connections are needed, it will open a connection, and then immediately close the connection upon it being returned to the pool. If fewer connections are needed, the unused connections are still held open forever, using resources on the DB server unnecessarily.

This commit adds a caching pool class that will keep returned connections open for a (user-configurable) period of time, enabling connection re-use and reducing connect/disconnect overhead during periods of heavy use, while freeing up resources by closing old connections during periods of lighter use, and removing the need to figure out a "magic number" of minion's to balance overhead and server resource use.

### DIFF
--- a/doc/src/pool.rst
+++ b/doc/src/pool.rst
@@ -62,3 +62,6 @@ be used.
         This pool class is mostly designed to interact with Zope and probably
         not useful in generic applications.
 
+.. autoclass:: CachingConnectionPool
+
+    .. note:: Expired connections are cleaned up on any call to putconn.
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -243,24 +243,24 @@ class PersistentConnectionPool(AbstractConnectionPool):
 class CachingConnectionPool(AbstractConnectionPool):
     """A connection pool that works with the threading module and caches connections"""
 
-    # A dictionary to hold connection ID's and when they should be removed from the pool
-    # Keys are id(connection) and vlaues are expiration time
-    # Storing the expiration time on the connection itself might be preferable, if possible.
-    from collections import OrderedDict
-    _expirations = OrderedDict()
-
     def __init__(self, minconn, maxconn, lifetime = 3600, *args, **kwargs):
         """Initialize the threading lock."""
         import threading
-        from datetime import datetime, timedelta
         AbstractConnectionPool.__init__(
             self, minconn, maxconn, *args, **kwargs)
         self._lock = threading.Lock()
         self._lifetime = lifetime
 
+        # A dictionary to hold connection ID's and when they should be removed from the pool
+        # Keys are id(connection) and vlaues are expiration time
+        # Storing the expiration time on the connection itself might be preferable, if possible.
+        from collections import OrderedDict
+        self._expirations = OrderedDict()
+
     def _connect(self, key=None):
         """Create a new connection, assign it to 'key' if not None,
         And assign an expiration time"""
+        from datetime import datetime, timedelta
         conn = psycopg2.connect(*self._args, **self._kwargs)
         if key is not None:
             self._used[key] = conn
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -370,13 +370,13 @@ class CachingConnectionPool(AbstractConnectionPool):
                 for conn in self._used.values():
                     if id(conn) == obj_id:
                         break  #found it, so just move on. Don't expire the
-                    # connection till we are done with it.
-                    else:
-                        # This connection doesn't exist any more, so get rid
-                        # of the reference to the expiration.
-                        # Can't delete here because we'd be changing the item
-                        # we are itterating over.
-                        junk_expirations.append(obj_id)
+                                # connection till we are done with it.
+                else:
+                    # This connection doesn't exist any more, so get rid
+                    # of the reference to the expiration.
+                    # Can't delete here because we'd be changing the item
+                    # we are itterating over.
+                    junk_expirations.append(obj_id)
 
             # Delete connection from pool if expired
             if del_idx is not None:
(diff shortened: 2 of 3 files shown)


dvarrazzo:
Thank you for your contribution. As mentioned in #563, would you mind having a chat in the ML about this change?

I'm sure a lot of people would benefit of a more useful pool implementation; I haven't used the pool object very often myself, so I'm not the best person to assess if your design is the best thing to do (rest assured that I find it an improvement).

Would it also be possible to have pool-related tests in the test suite? I don't think there is any.

ibrewster:
Yeah, I noticed the lack of tests. It was slightly disappointing, as I was planning to just copy and modify the existing tests for the ThreadedConnectionPool object for this one, but there wasn't one. I can look into what it would take to add some tests.

I have posted on the main python mailing list about this (in fact, I did that before I ever posted the original comment). Only one person there commented, suggesting I just up the minconn parameter. It looks like the mailing list link on the psycopg2 page links to the PostgreSQL mailing list, I could certainly post there as well.

dvarrazzo:
Docs need update too :)

ibrewster:
You mean in general for the pool classes or just in relation to this addition? :) I've never been terribly good at documenting stuff. My boss harps on that every few months :P

dvarrazzo:
Just this class: http://initd.org/psycopg/docs/pool.html

The page is a mix of what introspected from the docstring and docs-specific material, if that is more fitting to the docs than to the docstring: see [doc/src/pool.rst](https://raw.githubusercontent.com/psycopg/psycopg2/master/doc/src/pool.rst).

To generate the docs should be enough to run:

    cd doc
    make env
    make html

ibrewster:
Well, the ML discussion doesn't appear to be generating much interest so far. Maybe I'm the only one concerned about connection overhead :)

When I was writing the tests, I noticed that some of the utility test base classes provided didn't seem to be python3 compatible. Am I just missing something?

dvarrazzo:
All the psycopg code and tests is written for python2 and converted to python3 by 2to3 (as it was customary when py3 support was added). you can use `make PYTHON=python3` to convert the code and `make PYTHON=python3 check` to run the tests with a different Python interpreter (which results in a single command that you can run from command line anyway).

floqqi:
@ibrewster @dvarrazzo any progress here? Would love to see this in the new release :)

ibrewster:
Code is written and tested. I put it out on the mailing list for comment, but never really got any feedback.

zackw:
I have a use for this code in a web application that alternates between brief periods of extremely high load (so we don't want to be starting a fresh connection for every request) and multi-week periods of no load at all (during which time the database server processes should not hang around and tie up resources needed for other activity on the shared host).

So, with my end-user-of-psycopg2 hat on, I would like to vote for the feature; but then I would also like to offer some feedback on the implementation.  First off, it seems like expiry should be added at the level of `AbstractConnectionPool` so that it is available in all of the concrete pool classes; to put it another way, one could reasonably want this feature in the non-thread-safe pool as well as the thread-safe pool (maybe you're doing concurrency 100% with async/await).  Relatedly, it looks like the code for `CachingConnectionPool` repeats a lot of the `ThreadedConnectionPool` logic, can you please try to factor that out?  And finally, it doesn't seem sufficient to me to do cleanup only when a connection is released, because at that point you don't know if any more connections are going to come in before the expiration interval runs out.  Shouldn't there be some sort of timer that gets started on putconn, and reset on getconn, and only if the timer runs out do you call prune?

amachanic:
I just came here searching for a better connection pool than the default ones available, and discovered this rather old PR. Why was it never accepted?

Having spent a lot of time in the .NET world, I expect connection pools to work slightly differently than the Psycopg implementation. In ADO.NET the pool starts at the configured min size, and grows to the configured max size -- but along the way new connections are never closed. The pool keeps them, operating under the assumption that if at some point a high water mark is hit, it will probably be hit again. I think that's a much better default than the existing state of affairs in Psycopg.

That said, I really like the basic idea behind this PR. It is based on a much more sensible assumption than that of the built in pools, and can achieve an even better balance than the ADO.NET behavior. But I think the timing of the cache check is a bit weird (to @zackw's point). It seems to me that when I really want expiration is at some time after a spike, when things have cooled off. Consider that something kicked in, my program used 10 connections, and now it's an hour later and all is quiet. I'd like those connections to go away. I think that it makes sense, therefore, to launch a background thread that will wake up from time to time to do the appropriate work.

ibrewster:
The expiration method I wrote was done for simplicity and because it works well enough for my use case. I can certainly agree that it could be improved, however, to make it more useful for a wider variety of uses. That said, given that this PR has seen virtually no activity, it makes it harder to be motivated to improve on the code :)

I also agree with the points on refactoring - this module was based heavily off the ThreadedConnectionPool logic, and so it does share a lot of code. I didn't want to mess with any of the existing code, however, since it isn't mine, so I wrote it to be completely independent. That's something that could be changed if the authors of psycopg2 wanted it to be.

ibrewster:
pgpool/pgbouncer actually don't address the issue for which I wrote this pool: the need to open/close many connections from the application, and the overhead involved with that. They'll reduce the load on the server, sure, as well as providing connecting queuing and load balancing, if desired, but the application still has to open and close connections to them just as often. So unless the overhead of connecting to one of them is significantly less than the overhead of connecting directly to the DB, they don't help one iota in improving performance of an application that makes many short-lived connections :)

That said, I do agree with your point about the (apparent) lack of interest. Aside from the comments this PR has received, no one seemed interested - at least, not enough to comment on the ML. I can try again though, see if anything comes of it this time. Or I might just go ahead and release it separately as you mentioned. Might be the easiest option - that way it is available if anyone wants it, but the psycopg2 code base doesn't get cluttered with unused modules.

Changaco:
I'm interested in this. I didn't realize that the existing pool classes were flawed until I saw this PR, but  now that I know about it I'm guessing that it has a measurable impact on latency.

Changaco:
I'm going to open a new PR with a rebased and refactored version of the code in this branch.

Changaco:
The new PR is ready for review: #974.

dvarrazzo:
The psycopg2 pool probably will not receive further improvements. The [psycopg3 pool](https://www.psycopg.org/psycopg3/docs/api/pool.html) behaves in ways compatible with what you kindly suggested and implemented here: see [the description here](https://www.psycopg.org/psycopg3/docs/advanced/pool.html#pool-connection-and-sizing).

PR #566, opened 2017-06-09
Remove VC9 Runtime dependency and update appveyor.yml to newer versions

### LINKED ISSUE #541: On windows it always depends on vc9 runtime
I'm building python with vc2015 as well as psycopg but the setup.py script make it use a vc9 manifest unconditionally.
That makes it that even if it doesn't really need that runtime, the psycopg module cannot be loaded if you don't have the vc2008 redistributables.

jerickso:
In setup.py, changed the manifest check to only add the manifest when compiling with VC9 compiler instead of >= VC9.  Fixes #541 

Updated appveyor.yml to link against newer versions of OpenSSL and PostgreSQL.

Updated the runtime server version so the tests are now run against a PostgreSQL 9.6 server.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -73,7 +73,7 @@ matrix:
     fast_finish: false
 
 services:
-    - postgresql95
+    - postgresql96
 
 cache:
   # Rebuild cache if following file changes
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Thank you very much @jerickso ! Let's see if the OP has any feedback; I'll merge for 2.7.2 otherwise.

thierryba:
Sorry for the late reply. It is very likely to fix my issue. I'll report it again if that is not true. Thanks a lot.

PR #606, opened 2017-10-26
Emulate timeradd and timersub on Solaris

mykarlsson:
Solaris does not have timeradd and timersub. Add solaris_support.c which
provides emulated versions of them on Solaris.

### DIFF
--- a/psycopg/pqpath.c
+++ b/psycopg/pqpath.c
@@ -52,6 +52,10 @@
 #include "win32_support.h"
 #endif
 
+#if defined(__sun) && defined(__SVR4)
+#include "solaris_support.h"
+#endif
+
 extern HIDDEN PyObject *psyco_DescriptionType;
 extern HIDDEN const char *srv_isolevels[];
 extern HIDDEN const char *srv_readonly[];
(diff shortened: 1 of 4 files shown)


asheidan:
NICE!

dvarrazzo:
Perfect, thank you very much!

Do you know if there is any CI provider so that we can test for regression? I write the guys at http://polarhome.com/ but they never aswered...

mykarlsson:
None that I'm aware of unfortunately.

dvarrazzo:
In this case I'm afraid the QA level for solaris will remain "when something break, and then someone notices it, and then someone fixes it, then sooner or later that thing will be fixed". :)

dn0:
Hi @mykarlsson This patch broke psycopg2 on illumos and probably also on Solaris 11. Can you please test the fix mgerdts/psycopg2@eb711c7 on Solaris 10? (discussed in https://github.com/psycopg/psycopg2/issues/677#issuecomment-366439386)

mykarlsson:
Hi @dn0 absolutely, I'll reply under that issue.

PR #616, opened 2017-11-21
Use modern except syntax throughout project

jdufresne:
The syntax `except Exception, exc:` is deprecated. All Python versions supported by psycopg2 support the newer, modern syntax. Forward compatible with future Python versions.


dvarrazzo:
Makes sense, thank you.

PR #617, opened 2017-11-22
Use modern unittest TestCase assert methods

jdufresne:
Avoid the use of deprecated aliases. For example, `assertEquals()`, `failUnless()`, and `assert_()`. The newer common and idiomatic names are available on all Python versions supported by the psycopg2 library.

Helps keep the library in line with upstream idioms and keeps psycopg2's use of unittest more future proof.

Helps avoid deprecation warnings when Python is run with warnings enabled.

For a complete list of deprecated `TestCase` aliases, see:

https://docs.python.org/3/library/unittest.html#deprecated-aliases

### DIFF
--- a/tests/dbapi20.py
+++ b/tests/dbapi20.py
@@ -171,7 +171,7 @@ class DatabaseAPI20Test(unittest.TestCase):
             # Must exist
             threadsafety = self.driver.threadsafety
             # Must be a valid value
-            self.failUnless(threadsafety in (0,1,2,3))
+            self.assertTrue(threadsafety in (0,1,2,3))
         except AttributeError:
             self.fail("Driver doesn't define threadsafety")
 
@@ -180,7 +180,7 @@ class DatabaseAPI20Test(unittest.TestCase):
             # Must exist
             paramstyle = self.driver.paramstyle
             # Must be a valid value
-            self.failUnless(paramstyle in (
+            self.assertTrue(paramstyle in (
                 'qmark','numeric','named','format','pyformat'
                 ))
         except AttributeError:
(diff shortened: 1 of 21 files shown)


dvarrazzo:
I know what are the deprecated aliases, and I'm perfectly fine with them. Especially `assert_`: you will never catch me red-handed typing `assertTrue`.

More specifically, https://bugs.python.org/issue9424 was a really depressing show to assist, a public demonstration of stupidity and muscle flexing and the blatant demonstration that narrow-minded people who can't tell the difference between "consistency" and "foolish consistency" are dangerous, especially when they have commit privileges.

I prefer to keep this ode to stupidity to silence the deprecation warnings:

https://github.com/psycopg/psycopg2/blob/390e43f/tests/testutils.py#L67-L76

and IMO the only way to improve psycopg from the point of view of tests future-proofness is to switch to py.test.

jdufresne:
Thanks for reading and responding to the change request.

I think for new contributors, there is value in using methods that are documented in the official Python documentation. Additionally, I think there is value in using idioms and patterns common and established within the wider community. That is all I'm trying to achieve here, a project that is friendly to new contributors by using consistent Python patterns. Instead, the code is going out of its way to use methods documented as deprecated.

> I prefer to keep this ode to stupidity to silence the deprecation warnings:

This all seems unnecessary if the documented Python methods were used instead.

> IMO the only way to improve psycopg from the point of view of tests future-proofness is to switch to py.test.

I think py.test is great and think it would be great for psycopg2 to adopt it. However, that is a _much bigger_ and harder change. Even if that is the eventual goal, I see this as an incremental improvement.

PR #619, opened 2017-11-26
Add trove classifier for general Python 2 support

jdufresne:


dvarrazzo:
What is the python 2 trove classifier used for?

jdufresne:
It documents the libraries support. By adding this, The library says that is supports Python 2 in some capacity. This helps library users know, at a glance, if the library is compatible with a given project. I find this form of documentation much faster than searching docs, as PyPI always provides it in the same place.

Further, PyPI serves these values as a part of its JSON response. Scripts use these values to collect data, make decisions, and generate reports.

Thanks.

dvarrazzo:
Any link to the documentation of what the `Python :: 2` classifier is used for?

I added the `Python :: 3` because there was some web page on PyPI that didn't detect py3 support, see 4547753df0cc7819baff4420b27c381a7e221a8a. Is there any page/service around broken because `Python :: 2` is missing (although `:: 2.6` and `:: 2.7` are there)?

dvarrazzo:
`Programming Language :: Python :: Implementation :: CPython` should be probably added too (to differentiate it from pure python implementations using FFI with libpq).

Looking at the title of the commit linked above (Added explicit Python 2 and Python 3 classifiers) I suspect I tried adding `Python :: 2` at time and it didn't work: it might have not existed then but I am not sure.

Anyway, for consistency, I'll merge this request, but I don't think in practical term it does anything.

jdufresne:
Here is a complete list of all trove classifiers:

https://pypi.python.org/pypi?%3Aaction=list_classifiers

I don't think there is any specific documentation of their precise semantic meaning.

> Programming Language :: Python :: Implementation :: CPython should be probably added too

Added in the latest revision.

dvarrazzo:
> Here is a complete list of all trove classifiers:
>
> https://pypi.python.org/pypi?%3Aaction=list_classifiers

I know the list. I wonder if there is a svn/hg/git log for it... (I don't think it's part of the cpython codebase, might be pypi's)

jdufresne:
Best I could find, quickly:

https://github.com/pypa/pypi-legacy/commit/2df397d495622e77c260d8053cc217041489f53c#diff-137ce9fb4d42af2931c34e7dc4078ff9R2222

PR #625, opened 2017-11-27
Remove workaround for decimal module

jdufresne:
The decimal module is available on all Python versions supported by psycopg2. It has been available since Python 2.4. No need to catch an `ImportError`.

https://docs.python.org/2/library/decimal.html


dvarrazzo:
Thank you for this string of code cleanup MRs: I'll merge everything into master.

PR #629, opened 2017-11-28
Always prefer a dict literal over dict()

jdufresne:
Literals are a more idiomatic style and always faster.

Admittedly, the performance is a micro-optimization and doesn't matter for these tests, but as a matter of style and habit, always use a literal.

For completeness:

$ python3 -m timeit 'dict(a=1, b=2, c=3)'
10000000 loops, best of 3: 0.188 usec per loop
$ python3 -m timeit '{"a": 1, "b": 2, "c": 3}'
10000000 loops, best of 3: 0.0693 usec per loop

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -340,7 +340,7 @@ class ParseDsnTestCase(ConnectingTestCase):
 
         self.assertEqual(
             ext.parse_dsn('dbname=test user=tester password=secret'),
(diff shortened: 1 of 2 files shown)


dvarrazzo:
In the code that'd be fine, you could make a point of microoptimization, but in the tests I prefer clarity, and I find using `dict()` and keyword cleaner as it doesn't have the noise of the quotes.

PR #639, opened 2017-12-02
Trim trailing whitespace from all files throughout project

jdufresne:
Many editors automatically trim whitespace on save. By trimming all  files in one go, makes future diffs cleaner without extraneous whitespace changes.


psycopg2.sln
```diff
@@ -26,7 +25,7 @@ Global
 		$2.DirectoryNamespaceAssociation = None
 		$2.ResourceNamePolicy = FileName
 		$0.StandardHeader = $3
-		$3.Text = 
+		$3.Text =
```
dvarrazzo: I guess this crap is autogenerated: not worth fighting against it.

fogzot: sln is an artifact from using old MD ide for C/C++. Imho can be removed.

dvarrazzo: Good solution :)

PR #640, opened 2017-12-02
Remove sorted() workaround; available on all supported Pythons

jdufresne:
Introduced in Python 2.4 and the oldest supported Python is 2.7.

https://docs.python.org/2/library/functions.html#sorted

### DIFF
--- a/examples/dialtone.py
+++ b/examples/dialtone.py
@@ -17,13 +17,6 @@ from datetime import datetime
 import psycopg2
 from psycopg2.extensions import adapt, register_adapter
 
-try:
-    sorted()
-except:
-    def sorted(seq):
-        seq.sort()
-        return seq
-
 # Here is the adapter for every object that we may ever need to 
 # insert in the database. It receives the original object and does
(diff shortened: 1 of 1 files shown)


examples/dialtone.py
```diff
@@ -17,13 +17,6 @@
 import psycopg2
 from psycopg2.extensions import adapt, register_adapter
 
-try:
-    sorted()
```
dvarrazzo: Ha ha, this is also broken by the way :D `sorted()` always gives an error: missing argument :P It should have been `sorted` and catching `ImportError`...

Ok, these directories are largely archaeology anyway: I guess we are leaving them here for sentimental reasons...

jdufresne:
> Ha ha, this is also broken by the way :D sorted() always gives an error: missing argument :P It should have been sorted and catching NameError...

Ha. True.

If this directory of examples is not maintained and unverified, it may be worth considering removing it. If are people trying to use these examples and they don't work, it would lead to confusion.

PR #642, opened 2017-12-03
Officially deprecate psycopg1 module

jdufresne:
Per the module's comment, using `psycopg1` is considered a temporary hack that should not be relied on long term. Import now reports a warning. The warning instructs users to use the modern module instead. Allows
removing the workaround in a future version.

The `psycopg1` module has been available since the very first psycopg2 release. There has been plenty of time for users to migrate.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,7 +7,8 @@ What's new in psycopg 2.8
 Other changes:
 
 - Dropped support for Python 2.6, 3.2, 3.3.
-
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Nah. Let's just kill this module.

dvarrazzo:
Dropping the module in a separate branch.

jdufresne:
Even better!

Issue #558, opened 2017-05-23
psycopg2 incorrectly casting intervals from Redshift


tjhiggins:
This doesn't appear to be an issue with standard postgresql since I tried to replicate the scenario by adding a new test case, but it passed. Only an issue in Redshift as far as I know. 

```sql 
SELECT interval '1 day'
``` 
Correctly = `timedelta(days=1)`

```sql 
SELECT interval '1 day' from {table}
``` 
Incorrectly = `timedelta(days=0)` in psycopg2==2.6.1
It throws an overflow exception in psycopg2==2.7.1 

For whatever reason the psycopg2.extensions.INTERVAL value is `1 day` without selecting from a table, but if you select from any table the value is 1 day in microseconds `8.64e+10`. 

My current hack which probably wrong in other ways was to add a new_type to override the INTERVAL casting:
```python
def cast_redshift_interval(value, curs, _interval=psycopg2.extensions.INTERVAL):
    # https://github.com/psycopg/psycopg2/issues/558
    try:
        casted_value = _interval(value, curs)
    except OverflowError as error:
        # psycopg2==2.7.X incorrectly throws an OverflowError if you select an interval from a table
        try:
            return timedelta(microseconds=float(value))
        except ValueError:
            raise error

    # psycopg2==2.6.X incorrectly zeros out intervals if you select an interval from a table
    if isinstance(casted_value, timedelta) and casted_value.total_seconds() == 0:
        try:
            return timedelta(microseconds=float(value))
        except ValueError:
            return casted_value
    else:
        return casted_value

...

redshift_interval = psycopg2.extensions.new_type(
    psycopg2.extensions.INTERVAL.values,
    str('REDSHIFTINTERVAL'),
    cast_redshift_interval)
psycopg2.extensions.register_type(redshift_interval, conn)
```

dvarrazzo:
Hi, sorry, I didn't really notice this bug appearing.

What is the text representation of the case that fails in psycopg 2.7.1? I.e. if you run:

```
SELECT (interval '1 day')::text from {table}
```

what is the result?

Thank you: if we get a response quickly enough I'll try to fix this and release in a few day. Tentatively added to milestone 2.7.2.

tjhiggins:
@dvarrazzo Interestingly `SELECT (interval '1 day')::text from {table}` returns `1 day`.
However, `SELECT interval '1 day' from {table}` returns `86400000000`.
1 day in microseconds.

[...]

PR #658, opened 2018-01-02
Add Suport for msys2 enviroment

welitonfreitas:
this resolve erro on #558


dvarrazzo:
If you really want to declare msys2 supported you should add automatic build to appveyor. If you don't I can merge this MR (once the window breakage is fixed) but cannot guarantee it won't break again in future.

PR #727, opened 2018-06-06
Change DictRow representation to match that of dict

AlexCharlton:
I believe this more closely matches expectations of what the printed representation of this class should be.

### DIFF
--- a/lib/extras.py
+++ b/lib/extras.py
@@ -175,6 +175,9 @@ class DictRow(list):
             x = self._index[x]
         super(DictRow, self).__setitem__(x, v)
 
+    def __repr__(self):
+        return repr(dict(self))
+
     def items(self):
         g = super(DictRow, self).__getitem__
         return ((n, g(self._index[n])) for n in self._index)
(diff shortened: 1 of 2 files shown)


dvarrazzo:
What does it look, at the moment?

dvarrazzo:
`DictRow` is mainly a sequence, because it wants to maintain compatibility with the requirements of the DBAPI. `RealDictRow` is more dict than sequence, breaking something. So I'm kinda -0.5 about making this change...

AlexCharlton:
I certainly appreciate that `DictRow` acts as both a sequence and like a dict, and that is obviously fine. From a user perspective, however, you probably don't use `DictCursor` unless you want the dict part of it, and I think the representation of the object should reflect that fact and make it clear it's not returning regular rows.

I do recognize that this is a subjective matter, though :)

fogzot:
As @dvarrazzo said `DictRow` basically is a DBAPI compatible row that can *also* be accessed using column names. It makes sense to represent it as a list.

AlexCharlton:
¯\_(ツ)_/¯ as you wish

PR #728, opened 2018-06-10
Update intersphinx URLs to point to Python 3 docs

jdufresne:
Python 3 docs are more up to date and reflect the future of Python.

Removed unused py3 marker.


dvarrazzo:
cherry-picked

PR #730, opened 2018-06-15
Improve doc for extras.wait_select

blueyed:
btw: I assume this to be less performant - is that correct?
If so, a note about this could be added, too - pointing out that this should be used only in interactive mode and tests.

### DIFF
--- a/lib/extras.py
+++ b/lib/extras.py
@@ -759,8 +759,8 @@ def wait_select(conn):
 
     The function is an example of a wait callback to be registered with
     `~psycopg2.extensions.set_wait_callback()`. This function uses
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Cherry-picked

blueyed:
@dvarrazzo 
Thanks!

Can you comment on?

> btw: I assume this to be less performant - is that correct?

blueyed:
For reference: https://github.com/psycopg/psycopg2/commit/6af55ee51ca098850ccfc360fd3cde339af984f6

dvarrazzo:
I guess somewhat, as there is an extra C -> Python -> C every select, but I assume IO trumps CPU here.

PR #735, opened 2018-06-28
Add testing and document support for Python 3.7

jdufresne:
Python 3.7 was released on June 27, 2018.

https://docs.python.org/3/whatsnew/3.7.html

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,6 +21,9 @@ environment:
       - {PYVER: "35", PYTHON_ARCH: "64"}
       - {PYVER: "36", PYTHON_ARCH: "32"}
       - {PYVER: "36", PYTHON_ARCH: "64"}
+      - {PYVER: "37", PYTHON_ARCH: "32"}
+      - {PYVER: "37", PYTHON_ARCH: "64"}
+
 
     OPENSSL_VERSION: "1_0_2n"
     POSTGRES_VERSION: "10_1"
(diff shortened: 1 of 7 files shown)


dvarrazzo:
Gh! Apparently it's too early

jdufresne:
Ah! Perhaps I jumped the gun a bit. I've reverted the travis/appveyor changes so they'll continue to pass. They can be updated as soon as they adopt Python 3.7 support.

michael-k:
ref travis-ci/travis-ci#9815

.appveyor.yml
```diff
@@ -21,6 +21,9 @@ environment:
       - {PYVER: "35", PYTHON_ARCH: "64"}
       - {PYVER: "36", PYTHON_ARCH: "32"}
       - {PYVER: "36", PYTHON_ARCH: "64"}
+      - {PYVER: "37", PYTHON_ARCH: "32"}
+      - {PYVER: "37", PYTHON_ARCH: "64"}
+
```
dvarrazzo: There is a failure on appveyor: I think it requires at least

    - IF "%PYVER%"=="37" SET VS_VER=14.0

in the `init` stanza, and hope for the best. Can you try that please?

.travis.yml
```diff
@@ -4,12 +4,14 @@ dist: trusty
...
-  - 2.7
-  - 3.7-dev
-  - 3.6
-  - 3.5
-  - 3.4
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.7
+      dist: xenial
```
dvarrazzo: Looks like xenial doesn't support postgres 9.3 (and 9.2 I guess): I can help building them. If possible I'd move the entire grid on xenial.

jdufresne: > If possible I'd move the entire grid on xenial.

Sure thing. This will work for all versions except 3.4, which is not available on xenial.

dvarrazzo:
I've built the missing test database on xenial and all passed. Merged. Thank you very much :)

jdufresne:
Thanks!

Issue #714, opened 2018-05-18
Python 3.7 support: async is a keyword


hroncok:
According to PEP 492, `async` and `await` are now keywords and cannot be used as names.

psycopg2 does:

```pycon
  File "/usr/lib64/python3.7/site-packages/psycopg2/tests/test_async_keyword.py", line 43
    self.conn = self.connect(async=True)
                                 ^
SyntaxError: invalid syntax
```

```pycon
  File "/usr/lib64/python3.7/site-packages/psycopg2/tests/test_cursor.py", line 570
    connect_func = lambda: self.connect(async=True)
                                            ^
SyntaxError: invalid syntax
```

Can this be renamed, to `asynch` maybe or `async_`?

dvarrazzo:
It is already fixed: since psycopg 2.7 you can use `async_` instead. The tests should probably be fixed (at the moment Python 3.7 is not in travis) but the lib itself should work.

hroncok:
Running on Travis CI to see what happens BTW: https://travis-ci.org/hroncok/psycopg2/jobs/380610023

hroncok:
https://github.com/hroncok/psycopg2/commit/fcd97d05777b2da0257b9d1e79d16c4efce91079

hroncok:
Would you accept a PR that fixes the tests and enables 3.7 on Travis?

hroncok:
Funny enough the tests pass as they are.

dvarrazzo:
`test_async_keyword.py`, as the name says, is conditionally excluded from the test suite, see `tests/__init__.py`. `test_cursor.py` seems broken and can be changed using `async_`.

dvarrazzo:
Of course I'm happy to accept running the tests on 3.7 too, thank you :)

dvarrazzo:
Weird the tests pass as they are. I would have expected `test_cursor` to fail importing.

hroncok:
https://github.com/psycopg/psycopg2/commit/ef64493b8913e4069c4422ad14da6de405c445f6

dvarrazzo:
As of 2.7 the tests were passing on python 3.7. The cursor test was broken in psycopg 2.7.2 by b203a7c775c5f81f59c6b4205c3c33455ceeec3d

dvarrazzo:
> ef64493

this is in master but not in maint_2_7. Will cherry-pick.

hroncok:
[...]

PR #736, opened 2018-06-28
build fix: async is now a reserved keyword with Python3.7, rename it to async_

LocutusOfBorg:
Please review it carefully, it builds and installs fine now, but I didn't do any runtime test.

### DIFF
--- a/psycopg/connection.h
+++ b/psycopg/connection.h
@@ -96,7 +96,7 @@ struct connectionObject {
     int status;               /* status of the connection */
     xidObject *tpc_xid;       /* Transaction ID in two-phase commit */
 
-    long int async;           /* 1 means the connection is async */
+    long int async_;           /* 1 means the connection is async */
     int protocol;             /* protocol version */
     int server_version;       /* server version */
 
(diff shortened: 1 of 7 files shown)


dvarrazzo:
...And it breaks every other Python version.

What is this for? `async_` is already an alias for `async` since psycopg 2.7.

LocutusOfBorg:
I don't know this was the error I got, the package was not even installable with python3.7
```
Setting up python3-psycopg2 (2.7.5-1build1) ...
  File "/usr/lib/python3/dist-packages/psycopg2/tests/test_async_keyword.py", line 43
    self.conn = self.connect(async=True)
                                 ^
SyntaxError: invalid syntax

dpkg: error processing package python3-psycopg2 (--configure):
 installed python3-psycopg2 package post-installation script subprocess returned error exit status 1
```

dvarrazzo:
That file is not imported skipped in the test for python > 3.6. Apparently that's not enough. But the support for the alias `async_` is already in place.

dvarrazzo:
Psycopg 2.7.5 works fine on Python 3.7: tests can run with

    $ make PYTHON=/usr/local/py37/bin/python3
    $ make PYTHON=/usr/local/py37/bin/python3 check

I don't know what you are installing, it seems a `deb` package. We don't maintain those.

LocutusOfBorg:
So, you think we can just remove test_async_keyword.py file and live happy? this is not something that I'm hitting alone #714

PR #739, opened 2018-07-03
Remove setup.py fallback that assumes postgres 7.4.0.

benjaminp:
pyscopg2 only supports postgres >= 9.1 these days. Thus, taking this fallback will only lead to failure later down the line.


dvarrazzo:
Merged

PR #753, opened 2018-07-25
Allow unpacking a Range object like a tuple

jdufresne:
Define the `__getitem__` method on Range. Allows:

- Unpacking ranges as a `tuple`
- Indexing the `Range` with 0 & 1 like a `tuple`
- Coercing to a `tuple`
- Iterating like a `tuple`

### DIFF
--- a/lib/_range.py
+++ b/lib/_range.py
@@ -105,6 +105,10 @@ class Range(object):
             return False
         return self._bounds[1] == ']'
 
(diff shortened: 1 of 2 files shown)


jdufresne:
Just an alternative use to the existing API. I think it would allow for a useful construct such as:

```py
lower, upper = my_range
```
If you don't agree, no problem. We can close.

PR #755, opened 2018-07-26
Remove obsolete and incorrect FreeBSD version condition

AMDmi3:
The FreeBSD-related condition which enables custom `round()` implementation is incorrect: one must include `<sys/param.h>` to get `__FreeBSD_version` value, and since it's not included here, the check succeeds while it shouldn't. Before it worked somehow, but since python 3.7 it results in conflicting declarations of `round()`. The condition is also no longer needed since FreeBSD 5.3 is unsupported for 12 years.

This change thus fixes build of psycopg with python 3.7 on FreeBSD.

### DIFF
--- a/psycopg/config.h
+++ b/psycopg/config.h
@@ -154,8 +154,7 @@ typedef unsigned __int64    uint64_t;
 #endif
 
 /* what's this, we have no round function either? */
-#if (defined(__FreeBSD__) && __FreeBSD_version < 503000) \
-    || (defined(_WIN32) && !defined(__GNUC__)) \
+#if (defined(_WIN32) && !defined(__GNUC__)) \
     || (defined(sun) || defined(__sun__)) \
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you for this. Do you know if it's possible to set up a CI system for BSD?

AMDmi3:
If you mean public CI service like Travis, there are none I know of, unfortunately. Self hosting is always available with any opensource CI system, but maintaining this would be time wasting IMO. You could also just set a FreeBSD (or other BSD) VM and test your code there occasionally.

dvarrazzo:
@AMDmi3 self-hosting is not anything I fancy doing, no. As well as having my own VM. So this patch will be merged without further QA.

dvarrazzo:
@AMDmi3 can you please check if the MR #768 fixes the problem? Thank you

dvarrazzo:
merged

PR #768, opened 2018-09-05
Import params.h on FreeBSD if needed to establish version

dvarrazzo:
Fix #755


dvarrazzo:
Because I tend to not touch things I don't know of systems I don't own and can test on.

dvarrazzo:
Mmm...kay. I'll merge #755 instead.

Issue #773, opened 2018-09-21
Provide a better __str__ implementation for "range" types.


exhuma:
Currently converting the range types to strings, returns a `__repr__`. This is informative, but can be very lengthy in some cases. Sometimes (debugging, unit-tests, logging) a shorter string-form might be beneficial.

Implementing both `__str__` and `__repr__` would leave the choice of representation to the developer.

### Observed Behaviour

```
>>> from psycopg2.extras import DateTimeTZRange
>>> from datetime import datetime, timezone
>>> range = DateTimeTZRange(
...     datetime(1900, 1, 1, tzinfo=timezone.utc),
...     datetime(1900, 1, 10, tzinfo=timezone.utc))
>>> str(range)
"DateTimeTZRange(datetime.datetime(1900, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(1900, 1, 10, 0, 0, tzinfo=datetime.timezone.utc), '[)')"
>>> repr(range)
"DateTimeTZRange(datetime.datetime(1900, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(1900, 1, 10, 0, 0, tzinfo=datetime.timezone.utc), '[)')"
>>> repr(range) == str(range)
True
```

### Expected Behaviour

```
>>> from psycopg2.extras import DateTimeTZRange
>>> from datetime import datetime, timezone
>>> range = DateTimeTZRange(
...     datetime(1900, 1, 1, tzinfo=timezone.utc),
...     datetime(1900, 1, 10, tzinfo=timezone.utc))
>>> str(range)
"[1900-01-01 00:00:00+00:00, 1900-01-10 00:00:00+00:00)"
>>> # Or alternayively fully PG compliant syntax
>>> str(range)
'["1900-01-01 00:00:00+00:00", "1900-01-10 00:00:00+00:00")'
>>> repr(range)
"DateTimeTZRange(datetime.datetime(1900, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(1900, 1, 10, 0, 0, tzinfo=datetime.timezone.utc), '[)')"
>>> repr(range) == str(range)
False
```

dvarrazzo:
[...]

PR #779, opened 2018-09-30
Implement __str__ for range types

exhuma:
PR for #773

### DIFF
--- a/lib/_range.py
+++ b/lib/_range.py
@@ -62,6 +62,19 @@ class Range(object):
             return "%s(%r, %r, %r)" % (self.__class__.__name__,
                 self._lower, self._upper, self._bounds)
 
+    def __str__(self):
+        if (self._lower, self._upper, self._bounds) == (None, None, None):
+            return 'empty'
+
+        items = [
+            self._bounds[0],
+            str(self._lower),
+            ', ',
+            str(self._upper),
+            self._bounds[1]
+        ]
+        return ''.join(items)
+
     @property
     def lower(self):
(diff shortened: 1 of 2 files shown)


lib/_range.py
```diff
@@ -62,6 +62,19 @@ def __repr__(self):
             return "%s(%r, %r, %r)" % (self.__class__.__name__,
                 self._lower, self._upper, self._bounds)
 
+    def __str__(self):
+        if (self._lower, self._upper, self._bounds) == (None, None, None):
+            return 'empty'
```
dvarrazzo: I think this can be:

    if self._bound is None:
        return 'empty'

dvarrazzo: No worries, I'll fix this myself and merge :)

exhuma: Thanks :smile: 

It was my best guess after digging through the code.

dvarrazzo:
I have plans to work on psycopg 2.8 release in the next week. I won't promise it will be done though. I have just wrote to the ML with the plan for release and I have asked for help on a few open issues relatively easy to implement: see [my message](https://www.postgresql.org/message-id/CA%2Bmi_8auWr%2Bu4DwLGnu3JOu%3DjhMAV_8OPnMtMuXOr%2B0iMN5LgA%40mail.gmail.com).

Getting help on these open point (or working out there's too much work to do and postpone them) would be helpful. Otherwise for me psycopg2 is best-effort.

s22su:
@dvarrazzo any update, when this could be merged?

PR #800, opened 2018-10-30
Store WAL end pointer in the replication cursor

grunskis:
We have a couple of logical replication daemons running that are built using the psycopg2 library. They're replicating different databases with very different load patterns. One of these DBs has lots of load 24/7 and another one that has writes only during the office hours. We're closely monitoring the replication lag to avoid running out of the disk space on each of the DB instances.

I observed that the replication lag grows pretty fast during the night hours on the DB that has writes only during the day. I'm not 100% sure why the DB is growing when there are no writes, but I assume that's due to RDS or our Postgres instance configuration. In any case, we started getting replication lag notifications so I started digging into the problem.

I found that the `pg_stat_replication` view gives us the last LSN that was sent to the logical replication client. If we know this LSN and if we know that there were no changes made to the tables we're interested in we could just confirm from the client using the `send_feedback` call that the server can recycle the WAL files. Otherwise we have to wait until there's a write to the DB that will trigger a `ReplicationMessage` that contains the `data_start`.

But I couldn't find any way to get the current WAL end in the psycopg2 `ReplicationCursor`. So this PR adds the current WAL end pointer the client has received from the server with this specific replication cursor.

Let me know if this makes sense and/or you would like this to be changed in any way...

### DIFF
--- a/psycopg/pqpath.c
+++ b/psycopg/pqpath.c
@@ -1712,6 +1712,8 @@ retry:
         (*msg)->data_start = data_start;
         (*msg)->wal_end    = wal_end;
         (*msg)->send_time  = send_time;
+
+        repl->wal_end = wal_end;
     }
     else if (buffer[0] == 'k') {
         /* Primary keepalive message: msgtype(1), walEnd(8), sendTime(8), reply(1) */
@@ -1739,7 +1741,6 @@ retry:
         goto exit;
     }
 
-    repl->wal_end = wal_end;
     ret = 0;
 
 exit:
--- a/psycopg/pqpath.c
+++ b/psycopg/pqpath.c
@@ -1721,6 +1721,10 @@ retry:
             goto exit;
         }
 
+        wal_end = fe_recvint64(buffer + 1);
+        Dprintf("pq_read_replication_message: wal_end="XLOGFMTSTR, XLOGFMTARGS(wal_end));
+        repl->wal_end = wal_end;
+
         reply = buffer[hdr];
         if (reply && pq_send_replication_feedback(repl, 0) < 0) {
             goto exit;
@@ -1735,6 +1739,7 @@ retry:
         goto exit;
     }
 
+    repl->wal_end = wal_end;
     ret = 0;
 
 exit:
--- a/psycopg/replication_cursor.h
+++ b/psycopg/replication_cursor.h
@@ -44,6 +44,8 @@ typedef struct replicationCursorObject {
     struct timeval last_io;       /* timestamp of the last exchange with the server */
     struct timeval keepalive_interval;   /* interval for keepalive messages in replication mode */
 
+    XLogRecPtr  wal_end;          /* WAL end pointer from the last exchange with the server */
+
     XLogRecPtr  write_lsn;        /* LSNs for replication feedback messages */
     XLogRecPtr  flush_lsn;
     XLogRecPtr  apply_lsn;
--- a/psycopg/replication_cursor.h
+++ b/psycopg/replication_cursor.h
@@ -44,11 +44,11 @@ typedef struct replicationCursorObject {
     struct timeval last_io;       /* timestamp of the last exchange with the server */
     struct timeval keepalive_interval;   /* interval for keepalive messages in replication mode */
 
-    XLogRecPtr  wal_end;          /* WAL end pointer from the last exchange with the server */
-
     XLogRecPtr  write_lsn;        /* LSNs for replication feedback messages */
     XLogRecPtr  flush_lsn;
     XLogRecPtr  apply_lsn;
+
+    XLogRecPtr  wal_end;          /* WAL end pointer from the last exchange with the server */
 } replicationCursorObject;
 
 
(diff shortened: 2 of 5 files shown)


dvarrazzo:
I see you have changed the docs suggesting to use `wal_end`. But I understand from your description that using `flush_lsn=wal_end` is only safe if we know there is no change in the tables. What happens otherwise? Will the WAL files be recycled too early? If that's the case maybe this is a dangerous use and should be documented as a specific one (in `wal_end` docs) instead of in the generic example?

psycopg/replication_cursor.h
```diff
@@ -44,6 +44,8 @@ typedef struct replicationCursorObject {
     struct timeval last_io;       /* timestamp of the last exchange with the server */
     struct timeval keepalive_interval;   /* interval for keepalive messages in replication mode */
 
+    XLogRecPtr  wal_end;          /* WAL end pointer from the last exchange with the server */
+
```
dvarrazzo: Usually new members in a data structure should be added to the end

psycopg/replication_cursor_type.c
```diff
@@ -231,6 +231,17 @@ psyco_repl_curs_get_io_timestamp(replicationCursorObject *self)
...
 }
 
+/* object member list */
+
+#define OFFSETOF(x) offsetof(replicationCursorObject, x)
+
+static struct PyMemberDef replicationCursorObject_members[] = {
+    {"wal_end", T_ULONGLONG, OFFSETOF(wal_end), READONLY,
+        "LSN position of the current end of WAL on the server."},
+    {NULL}
```
dvarrazzo: Would there be value in adding attributes to access the other LSN members too?

grunskis: I thought about adding them, but decided not to b/c I didn't need them and wanted to keep the PR as small as possible.

grunskis:
> 
> 
> I see you have changed the docs suggesting to use `wal_end`. But I understand from your description that using `flush_lsn=wal_end` is only safe if we know there is no change in the tables. What happens otherwise? Will the WAL files be recycled too early? If that's the case maybe this is a dangerous use and should be documented as a specific one (in `wal_end` docs) instead of in the generic example?

I don't have good answers to all your questions which makes it a good reason to leave it out from the example and document it separately. I will do that...

My current understanding is that the server won't send keep-alive messages if there are changes to send and in that case we won't end up in the code that handles the socket timeout cases.

psycopg/pqpath.c
```diff
@@ -1735,6 +1739,7 @@ pq_read_replication_message(replicationCursorObject *repl, replicationMessageObj
         goto exit;
     }
 
+    repl->wal_end = wal_end;
```
dvarrazzo: Doing this here seems inconsistent. The variable may be uninitialized. Shouldn't it be done in the `buffer[0] == 'w'` block?

grunskis: I think you get here only from `buffer[0] == 'w'` block and only if there was no error there. I'll move it inside the block to be more clear.

dvarrazzo:
I'm sorry I can't be able to offer better guidance as of the safety of `flush_lsn=wal_end`: I'm not entirely practical with the replication subtleties. Maybe it's worth a chat to the `pgsql-hackers` to get better wisdom? I believe in the value of the MR you are offering: I'm just thinking about the better way to present it to the user and suggest them how to use it.

psycopg/replication_cursor_type.c
dvarrazzo: Makes sense to make it in a different branch, yes.

grunskis:
\

> 
> 
> I'm sorry I can't be able to offer better guidance as of the safety of `flush_lsn=wal_end`: I'm not entirely practical with the replication subtleties. Maybe it's worth a chat to the `pgsql-hackers` to get better wisdom? I believe in the value of the MR you are offering: I'm just thinking about the better way to present it to the user and suggest them how to use it.

I've removed the documentation change and added a warning instead.

I've looked into the source code of the "official" demo app https://github.com/postgres/postgres/blob/master/src/bin/pg_basebackup/pg_recvlogical.c

They basically do the same thing there. In the keep-alive message they include either the last LSN client has written to disk or the lastest walEnv value.

Sorry for taking too long to address the code review :)

gclinch:
Having access to the latest keep-alive message's wal_end would be a very useful addition for us too - can I help to progress this request?

grunskis:
I would love to see this merged ;)

We are running this in production for several months already in couple of projects w/o any issues.

@dvarrazzo Any tips on how to proceed to get closer to including this into one of the next versions?

dvarrazzo:
Merging to master to release in 2.8 :)

dvarrazzo:
Merged! Will be released with Psycopg 2.8 in a few days. Thank you for the feature and the testing, @grunskis!

gclinch:
@dvarrazzo I don't think the documentation changes in commit b8bf6d9917372d14967210f7a95ebb49fb25ecf3 correctly reflect this change.

The NEWS entry mentions Replication**Message**, but this PR is altering Replication**Cursor**.
Similarly extras.rst marks Replication**Message**.wal_end as new in 2.8, but it's Replication**Cursor**.wal_end that's new in 2.8

dvarrazzo:
@gclinch Oops, you are right. [fixed](file:///home/piro/src/psycopg2/doc/html/news.html#what-s-new-in-psycopg-2-8).

PR #813, opened 2018-11-21
add RETURNING clause support in execute_values function

eternalflow:
We found it useful in our project to improve `psycopg.extras.execute_values` in such a way that it supports `RETURNING` clause of query being executed.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,8 @@ New features:
   structure (:ticket:`#782`).
 - `~psycopg2.sql.Identifier` can represent qualified names in SQL composition
   (:ticket:`#732`).
+- Added *fetch* parameter to `~psycopg2.extras.execute_values()` function
+  (:ticket:`813`).
 - Fixed adaptation of numeric subclasses such as `IntEnum` (:ticket:`591`).
 - `!str()` on `~psycopg2.extras.Range` produces a human-readable representation
   (:ticket:`#773`).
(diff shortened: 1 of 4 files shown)


dvarrazzo:
It seems a good idea but it needs tests and docs at least

dvarrazzo:
You may want to call the parameter just `fetch`

setup.py
```diff
@@ -47,7 +47,7 @@
 # Take a look at https://www.python.org/dev/peps/pep-0440/
 # for a consistent versioning pattern.
 
-PSYCOPG_VERSION = '2.8.dev0'
+PSYCOPG_VERSION = '2.8.dev1'
```
dvarrazzo: This is not necessary, and doesn't belong to this branch, please remove and rebase

dvarrazzo:
Hello @eternalflow,

tests pass ok now, they weren't as of yesterday IIRC, so that's great.

Docs can be improved by stating that `fetch` was added in psycopg 2.8. This is better not done in the docstring but in `doc/src/extras.rst` in the `.. autofunction:: execute_values
` directive with a `versionchanged`. You can find many examples in the same file.

I will add a line to the NEWS file myself when we merge.

For the rest the merge request seems alright, thank you for your contribution.

setup.py
eternalflow: done

eternalflow:
@dvarrazzo Thank you for guidance
Is there any plan for release date?

PR #819, opened 2018-12-01
Remove unnecessary test decorator 'skip_if_cant_cast'

jdufresne:
ctypes is available and works on all supported Pythons. It has been
available since Python 2.5. The tests were written when Python 2.4 was
still supported.

### DIFF
--- a/tests/test_types_basic.py
+++ b/tests/test_types_basic.py
@@ -22,13 +22,14 @@
 # FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
 # License for more details.
 
+import ctypes
 import decimal
+import platform
 
 import sys
-from functools import wraps
 from . import testutils
 import unittest
-from .testutils import ConnectingTestCase, decorate_all_tests, long
+from .testutils import ConnectingTestCase, long
 
 import psycopg2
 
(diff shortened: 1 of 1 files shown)


jdufresne:
The `skip_if_windows` is added for the issue described in this SO post: https://stackoverflow.com/a/24575865.

WDYT? I suppose other alternatives would be be to find a different approach to testing `typecast_BINARY_cast` that is available on all platforms.

dvarrazzo:
I think I don't want to get into DLL hell on windows for a test. I can live with this test not running on Windows.

jdufresne:
After rebasing, the `skip_if_windows` decorator now works :slightly_smiling_face:

tests/test_types_basic.py
```diff
@@ -471,36 +471,14 @@ def getquoted(self):
...
-            return self.skipTest("can't test bytea parser: %s - %s"
-                % (self._exc.__class__.__name__, self._exc))
-
-        return f(self, *args, **kwargs)
-
-    return skip_if_cant_cast_
-
-
-@skip_if_cant_cast
+@skip_if_windows
```
dvarrazzo: As you taught us the arts of using `unittest.skipIf()` :grin: I think this would be better expressed with an:

    unittest.skipIf(
        platform.system() == 'Windows',
        "Not testing because we are useless with ctypes on Windows")

or something like that? `skip_if_windows` sounds like there is some feature which is not available on win, whereas this is only a test shortcoming

tests/test_types_basic.py
```diff
@@ -533,7 +511,7 @@ def test_blank_hex(self):
         rv = self.cast(b'\\x')
         self.assertEqual(rv, b'')
 
-    def test_full_hex(self, upper=False):
+    def _test_full_hex(self, upper=False):
```
dvarrazzo: Why was this needed? :)

jdufresne: This was from a previous version. You're right it is no longer necessary and has been reverted in the latest revision.

tests/test_types_basic.py
jdufresne: Done! :slightly_smiling_face:

dvarrazzo:
Hey, that's simple now :+1:

PR #823, opened 2018-12-02
Dropped deprecated PersistentConnectionPool

jdufresne:
This class was deprecated in 27cd6c4880161a715b1952be64cc96ebc968fa2b (Dec 2, 2012), which was first included in release 2.5. Enough time has passed for library uses to find an alternative solution.

This class was untested.


NEWS
```diff
@@ -237,7 +239,7 @@ What's new in psycopg 2.6.2
 - The `~psycopg2.extras.wait_select` callback allows interrupting a
   long-running query in an interactive shell using :kbd:`Ctrl-C`
   (:ticket:`#333`).
-- Fixed `!PersistentConnectionPool` on Python 3 (:ticket:`#348`).
+- Fixed ``PersistentConnectionPool`` on Python 3 (:ticket:`#348`).
```
dvarrazzo: Why the format change? I prefer to use the

    `!Thing`

format for Python identifiers and

    ``thing``

for generic literals.

jdufresne:
Fixed. Now uses `!Thing` syntax.

dvarrazzo:
Thank you very much for this and the other cleanups :) Waiting for CI to validate and merging :+1:

PR #828, opened 2018-12-07
mention postgresql 11 in install docs

wbolster:


yhoiseth:
Does this mean that the current release (2.7) supports PostgreSQL 11?

wbolster:
yes it does work

yhoiseth:
Thanks for the quick response. Also, I realized that I could check [the continuous integration configuration](https://github.com/psycopg/psycopg2/blob/0e89b9de2ca254b11c0221727a502522ab238fa5/scripts/travis_test.sh#L70), which shows that the tests are run against PosgreSQL 11:
```bash
# Postgres built from master
if [[ -n "$TEST_FUTURE" ]]; then
    run_test 11
fi
```

dvarrazzo:
Merged and deployed. http://initd.org/psycopg/docs/install.html#prerequisites

wbolster:
thanks!

PR #863, opened 2019-03-13
Convert `while 1:` statements to `while True:`

jdufresne:
A slightly more readable and modern syntax.

### DIFF
--- a/doc/src/advanced.rst
+++ b/doc/src/advanced.rst
@@ -12,7 +12,7 @@ More advanced topics
     conn.commit()
 
     def wait(conn):
-        while 1:
+        while True:
             state = conn.poll()
             if state == psycopg2.extensions.POLL_OK:
                 break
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Right. In ancient Python versions `while 1` was more efficient as it was constant-folded and optimised away:

```
dis.dis("""while 1: pass""")
  1           0 SETUP_LOOP               4 (to 6)
        >>    2 JUMP_ABSOLUTE            2
              4 POP_BLOCK
        >>    6 LOAD_CONST               0 (None)
              8 RETURN_VALUE
```

whereas `True` had to be fetched from globals to check it evaluated to true, so it resembled more:

```
dis.dis("""while something_true: pass""")
  1           0 SETUP_LOOP               8 (to 10)
        >>    2 LOAD_NAME                0 (something_true)
              4 POP_JUMP_IF_FALSE        8
              6 JUMP_ABSOLUTE            2
        >>    8 POP_BLOCK
        >>   10 LOAD_CONST               0 (None)
             12 RETURN_VALUE
```

I see it's not the case and `True` is as much efficient as 1, so will merge.

dvarrazzo:
cherry-picked

PR #865, opened 2019-03-13
Remove unused use_pg_dll from setup.py

jdufresne:
Unused since 3076046b3fa7ad6fe793315df543b887c4132a40.


dvarrazzo:
Sweet, we were using that 1 since 2006. So many memories...

Rerunning appveyor tests to check why they didn't pass.

dvarrazzo:
cherry-picked

PR #867, opened 2019-03-13
Use True/False instead of 1/0 to represent bool values

jdufresne:
Slightly more modern, readable, and Pythonic.


dvarrazzo:
cherry-picked

PR #868, opened 2019-03-13
Use unittest 'skip' feature to skip mxDateTimeTests

jdufresne:
Rather than deleting the class, use the skip feature. Provides a more
informative message during test output.

Never skip DatetimeTests as all supported Python environments have the
datetime module builtin.


dvarrazzo:
ok, cherry-picked this too. Thank you very much again for the cleanup!

PR #869, opened 2019-03-14
Simplify have_(py|mx)datetime logic in setup.py

jdufresne:
`have_pydatetime` is always `True`. Its value never changes. All supported Python environments have the `datetime` module.

The "psycopg requires a datetime module" error case could never be reached as the setup always falls back to using the Python` datetime` module if `mx.DateTime` doesn't exist.


jdufresne:
Obsoleted by #872

PR #870, opened 2019-03-14
Remove unnecessary `except ImportError` for Python datetime types

jdufresne:
The Python `datetime` module is available on all supported platforms. The types are always available and never raise an `ImportError`.


dvarrazzo:
cherry-picked

PR #872, opened 2019-03-16
Removed support for mxDateTime as the default date and time adapter

jdufresne:
The use_pydatetime configuration option has been removed. Now, always
default to Python's builtin datetime. mxDatetime support continues to be
available as an alternative to Python's builtin datetime.


dvarrazzo:
cherry-picked, with minimal change with the NEWS entry in a different position

PR #873, opened 2019-03-16
Remove unused import in test_errcodes

jdufresne:
`reload()` is a builtin function in Python 2.7 and `importlib.reload()` is available on 3.4+. The fallback to `imp.reload()` is never used.

### DIFF
--- a/tests/test_errcodes.py
+++ b/tests/test_errcodes.py
@@ -26,12 +26,11 @@ import unittest
 from .testutils import ConnectingTestCase, slow, reload
 
 try:
+    # Python 2.7
     reload
 except NameError:
-    try:
-        from importlib import reload
-    except ImportError:
-        from imp import reload
+    # Python 3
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Uhm... wouldn't be better to import reload into compat? Ok, that's for tests only though...

Throughout all the code there's ample use of `sys.version_info[0]`: I'd probably use the same pattern.

Even better I'd probably have `compat.PY2` and `PY3` to check against. 

I'll give a complete cleanup to the versioning check in the code (in the C code there is some mess I've recently seen...)

dvarrazzo:
Actually, `reload` here is just imported from `testutils` a few lines above :D so this import maze here can be just dropped.

jdufresne:
> Actually, reload here is just imported from testutils a few lines above :D so this import maze here can be just dropped.

Awesome, even better. Thanks!

PR #874, opened 2019-03-16
Move imports to the top of the module across tests

jdufresne:
Allows removing many duplicate imports and better follows PEP8 guidelines:

https://www.python.org/dev/peps/pep-0008/#imports

> Imports are always put at the top of the file, just after any module
> comments and docstrings, and before module globals and constants.


dvarrazzo:
Cherry-picked. There are a few other non-top imports, I'll take care of those.

PR #904, opened 2019-04-22
Windows build system rewritten in Python

dvarrazzo:
I've taken everything out of the `appveyor.yml` file and put it into a much more understandable Python script. It was a nice chance to play with Python 3.6 and other features I hadn't yet (f-strings, pathlib).

The appveyor file now looks like:

```yaml
build_script:
    - "%PYEXE% scripts\\appveyor.py build_script"

after_build:
    - "%PYEXE% scripts\\appveyor.py after_build"

before_test:
    - "%PYEXE% scripts\\appveyor.py before_test"

test_script:
    - "%PYEXE% scripts\\appveyor.py test_script"
```

which I think is a good result.

The script is used both by the psycopg2 repos to run the test and by the psycopg2-wheels to build the packages.


jerickso:
Looks good from my browsing of the code.

One issue I recall having before was figuring out where to set the postresql server config before the server started, hence my comment in the yaml code :).  But you have it in a good spot.

PR #913, opened 2019-05-06
Smart replication feedback

CyberDem0n:
This commit makes psycopg2 responsible for sending the status update (feedback) messages to the server regardless of whether a synchronous or asynchronous connection is used.

Feedback is sent every *status_interval* (default value is 10) seconds, which could be configured by passing a corresponding parameter to the `start_replication()` or `start_replication_expert()` methods.
The actual feedback message is sent by the `pq_read_replication_message()` when the *status_interval* timeout is reached.

The default behavior of the `send_feedback()` method is changed. It doesn't send a feedback message on every call anymore but just updates internal structures. There is still a way to *force* sending a message if *force* or *reply* parameters are set.

The new approach has certain advantages:
1. The client can simply call the `send_feedback()` for every processed message and the library will take care of not overwhelming the server. Actually, in the synchronous mode it is even mandatory to confirm every processed message.
2. The library tracks internally the pointer of the last received message which is not keepalive. If the client confirmed the last message and after that server sends only keepalives with increasing *wal_end*, the library can safely move forward *flush* position to the *wal_end* and later automatically report it to the server.

Reporting of the *wal_end* received from keepalive messages is very important. Not doing so casing:
1. Excessive disk usage, because the replication slot prevents from WAL being cleaned up.
2. The smart and fast shutdown of the server could last indefinitely because walsender waits until the client report *flush* position equal to the *wal_end*.

The new way of sending status update (feedback) messages is inspired by reading the source code of the `pg_recvlogical` and [pgjdbc](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/v3/replication/V3PGReplicationStream.java).

This implementation is only extending the existing API and therefore should not break any of the existing code.

### DIFF
--- a/psycopg/replication_cursor_type.c
+++ b/psycopg/replication_cursor_type.c
@@ -152,7 +152,7 @@ consume_stream(replicationCursorObject *self,
     CLEARPGRES(curs->pgres);
 
     self->consuming = 1;
-    if (keepalive_interval >= 1) {
+    if (keepalive_interval > 0) {
         set_status_interval(self, keepalive_interval);
     }
 
--- a/psycopg/replication_cursor_type.c
+++ b/psycopg/replication_cursor_type.c
@@ -96,19 +96,19 @@ exit:
 }
 
 #define consume_stream_doc \
-"consume_stream(consumer, keepalive_interval=10) -- Consume replication stream."
+"consume_stream(consumer, keepalive_interval=None) -- Consume replication stream."
 
 static PyObject *
 consume_stream(replicationCursorObject *self,
                                PyObject *args, PyObject *kwargs)
 {
     cursorObject *curs = &self->cur;
-    PyObject *consume = NULL, *res = NULL;
-    double keepalive_interval = 10;
+    PyObject *consume = NULL, *interval = NULL, *res = NULL;
+    double keepalive_interval = 0;
     static char *kwlist[] = {"consume", "keepalive_interval", NULL};
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|d", kwlist,
-                                     &consume, &keepalive_interval)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|O", kwlist,
+                                     &consume, &interval)) {
         return NULL;
     }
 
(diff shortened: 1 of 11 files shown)


dvarrazzo:
Thank you for the MR. It seems important enough that we could release the change in a bugfix release, if we are sure no code is broken by the change.

I would like to ask for feedback to @a1exsh and @ringerc, who are the authors of the original streaming replication support (#322)

psycopg/replication_cursor_type.c
```diff
@@ -124,8 +138,9 @@ consume_stream(replicationCursorObject *self,
     CLEARPGRES(curs->pgres);
 
     self->consuming = 1;
+    set_status_interval(self, keepalive_interval);
```
a1exsh: It looks like this is going to overwrite the value provided in the function to start the stream.  I wonder if we could make `None` the new default value for this parameter and then if the client didn't specify it explicitly, the value from starting the stream will be kept.

tests/test_replication.py
```diff
@@ -163,18 +163,20 @@ def test_start_and_recover_from_error(self):
...
         self.create_replication_slot(cur, output_plugin='test_decoding')
-
-        # try with invalid options
-        cur.start_replication(
-            slot_name=self.slot, options={'invalid_param': 'value'})
+        self.make_replication_events()
 
         def consume(msg):
-            pass
-        # we don't see the error from the server before we try to read the data
```
a1exsh: This was the behavior with older server versions, which is corrected in newer ones, e.g. v11.  No reason to test specific server behavior if we care about the client being able to recover from error.

a1exsh:
@dvarrazzo we've discussed the need for the change and specific approach taken (maintaining backwards compatibility was a priority) with @CyberDem0n before he has implemented it.

dvarrazzo:
> @dvarrazzo we've discussed the need for the change and specific approach taken (maintaining backwards compatibility was a priority) with @CyberDem0n before he has implemented it.

Thank you @a1exsh: really appreciated!

psycopg/replication_cursor_type.c
```diff
@@ -138,7 +152,9 @@ consume_stream(replicationCursorObject *self,
     CLEARPGRES(curs->pgres);
 
     self->consuming = 1;
-    set_status_interval(self, keepalive_interval);
+    if (keepalive_interval >= 1) {
```
dvarrazzo: probably better as `if (keepalive_interval > 0.0)` because being it >1 sec is a "business rule" but being it nonzero is "the user specified something" (which passed validation).

dvarrazzo:
Looks good to me. I will add a NEWS entry myself on merging. Is this in a mergeable state for you, @CyberDem0n, @a1exsh?

a1exsh:
Looks good to me!

Jan-M:
Thanks for the quick help and merging.

Issue #838, opened 2019-02-01
NamedTupleCursor is inefficient


Changaco:
Calling the standard library's `namedtuple` function over and over again to create ephemeral classes isn't efficient. (You may want to take a look at [`namedtuple`'s source code](https://github.com/python/cpython/blob/3.7/Lib/collections/__init__.py#L310-L312) if you're not familiar with it.)

Instead of doing that a single `Record` class could be created and used as the `row_factory`. It could be similar to `DictRow` but with a `__getattr__` method in addition to `__getitem__`.

The current `NamedTupleCursor` implementation also degrades performance by calling `re.compile()` and `re.sub()` repeatedly in its [`_make_nt()` method](https://github.com/psycopg/psycopg2/blob/3f20f7934ad3efdf793ae81a61a842bec7adb85c/lib/extras.py#L384-L398). The `Record` class could avoid that most of the time by only generating the python-compatible names when an attribute lookup fails.

Would a Pull Request implementing those changes be merged?

Changaco:
Yes. I also confirm that I was aware of that caching mechanism when I wrote my previous comment.

dvarrazzo:
So we are talking about a single `namedtuple()` call for every `execute()`, which implies a roundtrip to the server, the server running the query, returning the data on the network, and then there is all the Postgres to Python data type adaptation. To me it doesn't sound a worthwhile optimisation. Do you have numbers to back its need?

I'm more ok with micro-optimisations e.g. the `re.compile()` call, which are very probably not measurable either but at least they fit in a pattern of correct usage, and still produce a named tuple on a thing called `NamedTupleCursor`.

If we have to update the named tuple to "a thing with `__getattr__`" it should be an entirely different object (it could even be `DictRow` itself as far as I'm concerned), or calling it `NamedTupleCursor` would be a lie. The overhead of doing it is also in documentation and upgrade path for the users, and of course `NamedTupleCursor` should stay where it is and with unchanged semantics as a lot of users use it expecting precisely named tuples.

Changaco:
> Do you have numbers to back its need?

[...]

PR #928, opened 2019-06-04
Fix namedtuple cache

Changaco:
I've finally gotten around to testing @dvarrazzo's solution to #838. Unfortunately it doesn't fully work.

The problem is that the `lru_cache` function was used to wrap a non-static method, so it receives `self` as first argument instead of only receiving `key` as intended, and since `self` (the cursor) isn't always the same object, the cache misses.

The fix is a little akward because the `lru_cache` decorator doesn't play well with `classmethod` and `staticmethod` (it took me a while to find a way that works).

### DIFF
--- a/lib/extras.py
+++ b/lib/extras.py
@@ -372,10 +372,11 @@ class NamedTupleCursor(_cursor):
         key = tuple(d[0] for d in self.description) if self.description else ()
         return self._cached_make_nt(key)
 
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Thank you for playing with it even if don't agree with this approach: I really appreciate it.

I'll look into this correction and merge it. In the need to make a release, so it should be available in a few days

Issue #898, opened 2019-04-13
Adding password to dsn causes scripts to ignore CTRL-C


richardbann:
OS version: Ubuntu 18.04
psycopg2 version: 2.8.1

The following script can be stopped with CTRL-C:

```py
import psycopg2


while True:
    try:
        psycopg2.connect("host=postgres")
    except Exception as e:
        continue
    else:
        break
```

Adding a password causes the script to ignore CTRL-C:

```py
import psycopg2


while True:
    try:
        psycopg2.connect("host=postgres password=x")
    except Exception as e:
        continue
    else:
        break
```

Please help, what makes such a big difference?

dvarrazzo:
Funny, it's true.

The command in itself terminates on error (if the password is wrong):

```
In [4]: psycopg2.connect("host=localhost user=wat dbname=piro password=y")
---------------------------------------------------------------------------
OperationalError                          Traceback (most recent call last)
<ipython-input-4-f5b524a63be1> in <module>()
----> 1 psycopg2.connect("host=localhost user=wat dbname=piro password=y")

/home/piro/dev/psycopg2/build/lib.2.7/psycopg2/__init__.pyc in connect(dsn, connection_factory, cursor_factory, **kwargs)
    124 
    125     dsn = _ext.make_dsn(dsn, **kwargs)
--> 126     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
    127     if cursor_factory is not None:
    128         conn.cursor_factory = cursor_factory

OperationalError: FATAL:  password authentication failed for user "wat"
FATAL:  password authentication failed for user "wat"
```

But in a loop it doesn't receive a SIGINT:

```
In [5]: while True:
   ...:     try:
   ...:         psycopg2.connect("host=localhost user=wat dbname=piro password=y")
   ...:     except Exception as e:
   ...:         continue
   ...:     else:
   ...:         break
   ...:     
^C^C^C^CTerminated
```

(until killed with a TERM).

If I add a couple of prints, before the `connect()` and after the `except`, I can see the command in a tight loop, and pressing ctrl-C the output is "disturbed" by the `^C` chars appearing and streaming up. So it doesn't seem a psycopg bug, rather a Python one. Maybe the quick release and acquisition of the GIL doesn't allow the signal handler to operate correctly.

Anyway, if you put a `time.sleep(0.01)` after the `except` it will receive and handle the ctrl-c ok.

richardbann:
Thanks for looking into it.

[...]

PR #934, opened 2019-06-18
Handle exit exceptions in connect

richardbann:
I was trying to find the reason causing the strange behavior described in #898.
I understand this is a marginal case, but anyway it feels to me the code is "more correct" this way.
Please let me know if you wish to have a look.

PS.: Even if this will not be merged, the deeper understanding of the internals did worth it  :nerd_face:


dvarrazzo:
If you need to fudge too much with timing in the test it's not reliable enough: you should use other synchronisation methods.

psycopg/connection_type.c
```diff
@@ -1395,7 +1395,15 @@ connection_setup(connectionObject *self, const char *dsn, long int async)
         PyObject *ptype = NULL, *pvalue = NULL, *ptb = NULL;
         PyErr_Fetch(&ptype, &pvalue, &ptb);
         obscure_password(self);
-        PyErr_Restore(ptype, pvalue, ptb);
+        // if we got a system exit exception (not a subclass of Exception),
```
dvarrazzo: `//` comments are not used in psycopg codebase as per PEP 7

dvarrazzo:
Apart from the problem with the test, I'm generally fine with the proposed change. You are asking pretty much to replace a `try ... except:` with a `try ... except Exception` which is the right pattern.

psycopg/connection_type.c
richardbann: sure

richardbann:
I have no clue how to "raise KeyboardInterrupt" in the subprocess on windows. Maybe the windows case could be skipped in the tests?

dvarrazzo:
Quite frankly, this test is getting so complex and fraught with race condition that I'm not sure I want it in the codebase. And in general if window causes grief I'm ok with skipping the test.

I'm taking a closer look at the change proposed. How does it work? It seems that in case of Ctrl-C it _swallows_ the exception rather than ignoring it (as a `except Exception` would do).

Calling `PyErr_Fetch` without calling `PyErr_Restore` also makes me a bit nervous. I'll take a look at the cpython sources but [the docs](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch) seem adamant that they are designed to be used in a pair.

dvarrazzo:
Ah I think I see: the problem in #898 causes Ctrl-C to get handled into `obscure_password()`, so the `KeyboardInterrupt` is swallowed by the `PyErr_Restore()`. 

What you are doing is pretty much to force the opposite and to clobber the connection exception with a `KeyboardInterrupt`. I'm not sure I like it...

richardbann:
What I can think of is this:
```c
exit:
    /* here we obfuscate the password even if there was a connection error */
    {
        PyErr_CheckSignals(); /* at least we store and restore the KeyboardInterrupt */
        PyObject *ptype = NULL, *pvalue = NULL, *ptb = NULL;
        PyErr_Fetch(&ptype, &pvalue, &ptb);
        obscure_password(self);
        PyErr_Restore(ptype, pvalue, ptb);
    }
    return rv;
}
```
This also replaces the connection error with `KeyboardInterrupt` but I think having both is impossible.

dvarrazzo:
Handling the signal feels a better approach, yes.

Could you please try to move CheckSignal before connect if it solves your problem? If so I'd be sure the ctrl-c doesn't clobber a connection exception and would accept the patch alright.

richardbann:
I do not think moving CheckSignal before connect solve it. The root cause of the problem is that if CTRL-C is received **during** connect (in my original case during name resolution) python's signal handler is notified and during obscure_password (namely in the psyco_make_dsn call, in PyObject_Call) python sets the KeyboardInterrupt (I suspect PyObject_Call calls CheckSignal) which is than cleared by PyErr_Restore.

dvarrazzo:
Yes, what happens is that: this function is too C-tight so the first chance it has to manage signals is when it gets into a bit of Python machinery, that's all correct.

What I'm saying is that my line in the sand is that the connect function will not clobber a connection error with any less important exception, hence ignoring any error in obscure_password. This is really important for me. If there is an error shouting MITM ATTACK I don't want it to be hidden by the ctrl-c.

If  the ctrl-c happens to be caught just before the connection attempt I could make you happy and merge that. I can see how the connection is long so it's very likely that the signal is received in that time and processed afterwards.

In this case, nope. As already said in the initial bug report, looping so tightly around a connection attempt, without a logging, without a time delay, is just a broken way to handle a connection attempt and I'm not available to mask a connection error to allow it to work.

richardbann:
I think it was a mistake to demonstrate the problem with a tight loop. Let me go with a different approach.
If I run this script and press `CTRL-C` within 5 secs:
```python
import psycopg2

psycopg2.connect(host="10.255.255.1", connect_timeout=5)
print("done")
```
I get nothing, even if there was a timeout. `KeyboardInterrupt` "clobbered" the connection error. This is how it works now.

But, if I modify the code and run this script and press `CTRL-C` again (note the password kwarg)
```python
import psycopg2

psycopg2.connect(host="10.255.255.1", password="x", connect_timeout=5)
print("done")
```
I get  `psycopg2.OperationalError: timeout expired`.

What makes me feel a little bit weird is that adding a password changes the behavior fundamentally. Well, maybe not fundamentally but in an unexpected way.

To summarize:
- "connection error clobbering" happens with the existing code also, but in a (seemingly) unpredictable fashion
- it has nothing to do with the tight loop, and - at least to me - this is the expected behavior
- "making me happy" can not be a reason to change or merge anything, but it could be a side effect of making software better. I appreciate it though... :1st_place_medal: 
- If you are OK with the status quo I am fine to accept it, do not feel the need to convince me, you can close this PR

Thanks a lot.

dvarrazzo:
I understand that adding the password changes the code in such a surprising way, and I agree that surprising behaviours are a bad thing in coding. My point is that if the cure is worse than the problem I won't take it, but I'm keeping this MR open to check if we can solve the issue in an optimal way.

That pressing ctrl-c within 5 seconds doesn't do anything is not something that can be addressed: it's the libpq blocking. If any working in non-blocking fashion should handle that. However, when the 5 seconds expire, I would like that 100% of the times we exit with a timeout error, with or without password, and swallowing exceptions around `obscure_password()` accomplished that. I don't find acceptable to exit with KeyboardInterrupt: if you had really exited with ctrl-c you would have left before the 5 seconds timeout; you are only leaving our code because libpq made us the grace to leave, after 5 seconds, regradless the ctrl-c you pressed.

So if you are proposing something that does:

- connect(timeout=5)
- 2 seconds pass
- ctrl-c pressed
- 3 seconds passed
- KeyboardInterrupt

then I can tell you no, that's not a good behaviour, I wouldn't like it.

if what happens could be

- connect(timeout=5)
- 2 seconds pass
- ctrl-c pressed
- KeyboardInterrupt

then yes, let's see if we can get here. If it's impossible because libpq is a safe, then let's close it here.

dvarrazzo:
_ponder ponder..._

not that the swallowning ctrl-c altogether is a really really good behaviour either...

richardbann:
I guess your 2nd option is only doable by doing startconnect + poll even in the sync case. Nice puzzle...

dvarrazzo:
Of course we could move the password obfuscation before the connection, couldn't we?...

richardbann:
Yes, it would solve all my concerns but you get back 2s pass, ctrc-c, 3s pass, KeyboardInterrupt, which you did not like.

dvarrazzo:
I know. But at least it's not our fault...

If password obfuscation is moved before connection then the exception saving is not even needed anymore, whichi wouldn't mind at all.

richardbann:
I'm willing to try but will need hands-on support as I have no C experience at all. Where exactly that code should go?

dvarrazzo:
I would probably do something like f40ad0f3ae8f14ac4c88084eb1e201a94aa9759b

This drops the exception juggling that causes #898. This way I'd say we are doing the right thing as the function will raise a connection error AND we don't tamper with the signal handling and swallow a ctrl-c.

What do you think?

richardbann:
This is great! Thanks again!

dvarrazzo:
Thank you for the discussion. The issue has been fixed in master as explained above. Closing this merge request. Have a good day!

PR #935, opened 2019-06-18
Take libdir from pg_config --ldflags

mhaas-work:
On OS X with OpenSSL installed via brew, `-lssl` is not found
during the linker stage. This can be remedied by correctly taking into
account the build-time settings for PostgreSQL.

Since the format is incompatible, this commit introduces some string
parsing to retrieve the directories referenced in the linker call.

So, this commit fixes the build on OS X for me with PostgreSQL and OpenSSL installed from brew.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -409,6 +409,11 @@ For further information please check the 'doc/src/install.rst' file (also at
 
         try:
             self.library_dirs.append(pg_config_helper.query("libdir"))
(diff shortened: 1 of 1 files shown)


dvarrazzo:
@fogzot do you see anything obviously wrong with this approach?

dvarrazzo:
Thank you very much for this MR. I have adopted this approach, with some modifications, in the setup.py. This should probably solve #1200

PR #951, opened 2019-08-05
[psycopg/pqpath.c] sys/time.h include for non mac & Windows

SamuelMarks:
Resolving this error: https://travis-ci.org/glaucoma/glaucoma-analytics-rest-api/jobs/567715976#L1368

### DIFF
--- a/psycopg/pqpath.c
+++ b/psycopg/pqpath.c
@@ -51,10 +51,10 @@
 #include <winsock2.h>
 /* gettimeofday() */
 #include "win32_support.h"
-#endif
-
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Uhm... just to understand, how is that it seems needed on your platform and we didn't have any problem so far? Is it any particular compiler, platform, C library?

SamuelMarks:
@dvarrazzo Yeah, it's a weird one. Could be that CPython has a bunch of standard includes by default, whereas with PyPy—on LLVM—you need to explicitly include.

Issue #955, opened 2019-08-21
Why does get_backend_pid return a value different from pg_backend_pid()?


svetlyak40wt:
Here what I've got.

```
(Pdb) conn.get_backend_pid()
376976152

(Pdb) cur = conn.cursor()
(Pdb) cur.execute('SELECT pg_backend_pid()')
(Pdb) cur.fetchall()
[(699839,)]
```

### Environment

```
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.11 (jessie)
Release:	8.11
Codename:	jessie

psycopg2==2.7.5
postgresql-client-10==10.7-1.pgdg80+1
postgresql server==10.10
```

dvarrazzo:
Maybe you are connecting to a connection pool? Or a not-exactly-postgresql service?

The value returned by `conn.get_backend_pid()` is what returned at connection and retrieved using the libpq function [`PQbackendPID()`](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQBACKENDPID), psycopg won't perform an extra query. If it doesn't agree with `SELECT pg_backend_pid()` we cannot do much about it. The value in your example seems a bit fishy to be a unix process pid.

What server are you connecting to?

svetlyak40wt:
I'm connecting to a PGBouncer.

dvarrazzo:
PGBouncer seem to return a funny value on connection. You fixed your code alright avoiding the use of `conn.get_backend_pid()`.

svetlyak40wt:
Probably, a note about this unexpected behaviour should be added to the docs of get_backend_pid?

I could make a pull if you dont mind. What do you think, @dvarrazzo ?

svetlyak40wt:
Well, I've created a pull anyway. Hope it will be useful.

https://github.com/psycopg/psycopg2/pull/956

dvarrazzo:
Appreciated, thank you

PR #956, opened 2019-08-23
Added a note about undefined behavior of get_backend_pid when connecting to PgBouncer

svetlyak40wt:
This warning probably will save some time people, who are using proxies.

Original issue is: https://github.com/psycopg/psycopg2/issues/955


dvarrazzo:
The PGBouncer url is broken...

svetlyak40wt:
Good catch!

I've fixed the link!

PR #970, opened 2019-09-20
Add a new row type

Changaco:
This branch introduces a new row type: `HybridRow`. Unlike the other types, hybrid rows support both item and attribute lookups and assignments; and unlike namedtuples, hybrid rows are mutable and they're all instances of a single class.

### DIFF
--- a/doc/src/extras.rst
+++ b/doc/src/extras.rst
@@ -97,6 +97,10 @@ Real dictionary cursor
 `namedtuple` cursor
 ^^^^^^^^^^^^^^^^^^^^
 
+This type of cursor is especially well suited if you need to fetch and process a
+large number of similar rows at once, because tuples occupy less memory than
+dicts.
+
 .. versionadded:: 2.3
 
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Hi,

I am not really interested in adding an editable tuple type in the adapter.

If you want to champion such object please open a discussion in the mailing list.

Changaco:
No, I'm not going to join your mailing list. I opened a discussion by opening this PR, you chose to close that discussion. The obvious conclusion is that you probably have no intention of ever merging this branch.

Since the hybrid row type isn't wanted here I'll put it in postgres.py instead.

dvarrazzo:
This is a bug tracker, pretty much only under my eyes. New ideas ought to be discussed if they want to be included into an industry-standard project, if they are pushed to be used by all the world, if they have to be maintained by others.

It's good stuff that you take responsibility for your solely designed object instead of dumping it on us without discussion, yes. Put your class wherever you want.

PR #972, opened 2019-09-23
Suggest ujson.loads in documentation

FedericoCeratto:
Related to #341


dvarrazzo:
I will use a different wording but I'll mention the ujson example, ok.

PR #974, opened 2019-09-23
Implement keeping pooled connections open

Changaco:
This PR is a rebased and refactored version of #565.

When a connection is added (via `_connect()`) or returned (via `_putconn()`) to the pool, a timestamp is saved. When `_getconn()` pulls a connection, it compares the saved timestamp to the current time, and if the connection has been idle for longer than `idle_timeout`, then that connection is closed and discarded.

I've set the default value of `idle_timeout` to `0` in order to replicate the current behavior, however another value could be chosen if there is consensus on changing the default behavior.

My first commit fixes two bugs in the current pool implementation:

1. The `_getconn()` method doesn't check the status of a connection, so it can return a broken connection.
2. The `_putconn()` method doesn't reopen new connections if the total has dropped below `minconn`.

### DIFF
--- a/doc/src/pool.rst
+++ b/doc/src/pool.rst
@@ -12,13 +12,22 @@ Creating new PostgreSQL connections can be an expensive operation.  This
 module offers a few pure Python classes implementing simple connection pooling
 directly in the client application.
 
-.. class:: AbstractConnectionPool(minconn, maxconn, \*args, \*\*kwargs)
+.. class:: AbstractConnectionPool(minconn, maxconn, \*args, idle_timeout=0, \*\*kwargs)
 
     Base class implementing generic key-based pooling code.
 
     New *minconn* connections are created automatically. The pool will support
     a maximum of about *maxconn* connections.  *\*args* and *\*\*kwargs* are
     passed to the `~psycopg2.connect()` function.
+
+    Connections are kept in the pool for at most *idle_timeout* seconds. There
+    are two special values: zero means that connections are always immediately
+    closed upon their return to the pool; `None` means that connections are kept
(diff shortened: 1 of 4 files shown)


Changaco:
@dvarrazzo Please let me know if you would prefer that I release this as a separate package.

Changaco:
I've created a separate package: <https://pypi.org/project/psycopg2-pool/>.

dvarrazzo:
Hi, thank you for working on this issue, notwithstanding my lacklustre direct interest.

Bit of a background: I hate psycopg connection pool. I don't think it does the right things. Its origin probably comes from Zope and its footprint is still there - and it's likely wrong outside of it.

I think a good connection pool can be passive like psycopg one, if needed, but it might even do some smart, and have some form of internal keepalive and a better algorithm to gauge its size depending on its usage - in general be a bit bolder with its own behaviour. If done right it would probably break the psycopg pool interface, and I really see that module a bit of a legacy thing.

I think it would be a good idea to start from scratch and have a pool module of its own and a pool interface independent from psycopg pool, as well as having a release cycle independent from psycopg. I'm glad you did the first step on your own. If you would like collaboration for that module's design (breaking from bloody `_getconn/_putconn()`) I'd be happy to collaborate.

Cheers

-- Daniele

Changaco:
@dvarrazzo Feel free to open an issue in <https://github.com/Changaco/psycopg2-pool> so that we can discuss your ideas. I didn't start from scratch, but I did remove the weird legacy stuff from the API and implementation.

DzigaV:
@dvarrazzo Nominating something derived from this PR for psycopg3 🙂

Issue #990, opened 2019-10-18
Unable to install python 3.8 windows 10


tolstislon:
Windows 10 Python 3.8

```
(untitled1) D:\python_other\untitled1>pip install psycopg2-binary
Collecting psycopg2-binary
  Using cached https://files.pythonhosted.org/packages/80/91/91911be01869fa877135946f928ed0004e62044bdd876c1e0f12e1b5fb90/psycopg2-binary-2.8.3.tar.gz
    ERROR: Command errored out with exit status 1:
     command: 'c:\users\user\.virtualenvs\untitled1-lq2ited9\scripts\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\user\\AppData\\Local\\Temp\\pip-ins
tall-9839um0z\\psycopg2-binary\\setup.py'"'"'; __file__='"'"'C:\\Users\\user\\AppData\\Local\\Temp\\pip-install-9839um0z\\psycopg2-binary\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"
'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\user\AppData\Local\Temp\
pip-install-9839um0z\psycopg2-binary\pip-egg-info'
         cwd: C:\Users\user\AppData\Local\Temp\pip-install-9839um0z\psycopg2-binary\
    Complete output (23 lines):
    running egg_info
    creating C:\Users\user\AppData\Local\Temp\pip-install-9839um0z\psycopg2-binary\pip-egg-info\psycopg2_binary.egg-info
    writing C:\Users\user\AppData\Local\Temp\pip-install-9839um0z\psycopg2-binary\pip-egg-info\psycopg2_binary.egg-info\PKG-INFO
    writing dependency_links to C:\Users\user\AppData\Local\Temp\pip-install-9839um0z\psycopg2-binary\pip-egg-info\psycopg2_binary.egg-info\dependency_links.txt
    writing top-level names to C:\Users\user\AppData\Local\Temp\pip-install-9839um0z\psycopg2-binary\pip-egg-info\psycopg2_binary.egg-info\top_level.txt
[...]

PR #992, opened 2019-10-19
Add Python 3.8 to the test matrix

jdufresne:
Python 3.8 was released on October 14th, 2019.

- Added 'Programming Language :: Python :: 3.8' trove classifier.
- Added 'py38' to the tox test matrix.
- Added 'python: 3.8' to the Travis test matrix.
- Removed 'dist: xenial' from Travis configuration; it is now the
  default.
- Removed 'dist: trusty' from Travis configuration; it is not longer
  necessary.
- Removed 'sudo' from Travis configuration; it is deprecated.

https://docs.python.org/3.8/whatsnew/3.8.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,15 @@
 # Travis CI configuration file for psycopg2
 
-dist: xenial
-sudo: required
 language: python
 
 matrix:
   include:
     - python: 2.7
+    - python: 3.8
     - python: 3.7
     - python: 3.6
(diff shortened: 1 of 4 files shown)


dvarrazzo:
Thank you: was just about to work on this.

Will try to build on Windows too

dvarrazzo:
Appveyor doesn't support Python 3.8 at the moment so won't build that for now. (#990)

PR #1000, opened 2019-10-23
Drop support for EOL Python 3.4

### LINKED ISSUE #998: Drop Python 3.4 support
Python 3.4 has reached EOL in March, 2019.
Psycopg2 should drop support for it as well.

hugovk:
Fixes #998

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,8 +21,6 @@ environment:
         - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015, PY_VER: "36", PY_ARCH: "64"}
         - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015, PY_VER: "35", PY_ARCH: "32"}
         - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015, PY_VER: "35", PY_ARCH: "64"}
-        - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015, PY_VER: "34", PY_ARCH: "32"}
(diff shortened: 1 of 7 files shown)


hugovk:
Merge conflict resolved.

hugovk:
Merge conflict resolved.

dvarrazzo:
I was thinking that, yes.

I have just created a maint_2_8 branch: future 2.8.x releases will come from there, and we will not drop support for new python releases there.

If you can rebase this on master, we can merge it, then we can drop 3.5 and 2.7 in other changesets.

Documentation should be changed too: see Prerequisite section.

If I can suggest, let's first drop 3.5, which is probably only few changes similar to these, then dropping 2.7 is going to be much more of a butchering.

dvarrazzo:
Also missing from this changeset (and future ones) an entry in the NEWS file describing the versions dropped (see e.g. what's new in 2.8)

hugovk:
Rebased, prerequisites updated (also added 3.9) and added NEWS entry!

jerickso:
Removing Python 2.7 and 3.4 support will also help with moving the Windows builds to use the latest versions of Postgres libpq.

Basically, the Postgres 12 and 13 build scripts included in their distribution does not support Visual Studio versions before 2013 (Python 2.7 uses 2008 and Python 3.4 uses 2010).

I'll work on a pull request against master that will drop support for Python 2.7 and 3.4 for the Windows builds and use Postgres 13.

dvarrazzo:
I thought I had merged this already.... Sorry. Merged now, thank you very much!

hugovk:
No problem, you're welcome!

PR #1039, opened 2020-02-03
Emit a warning when a connection or cursor isn't closed

jdufresne:
This follows the semantics of Python file objects. When the object is garbage collected and it has not been closed, emit a `ResourceWarning`. This allows library users to notice inconsistent and non-deterministic resource management and fix it. Users should use context managers and finally blocks to always close cursors and connections when they are not longer required.
    
For example, in Python, not closing a file results in the following:
    
        $ python3 -Walways
        >>> f = open('foo', 'w')
        >>> del f
        __main__:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='foo' mode='w' encoding='UTF-8'>
        ResourceWarning: Enable tracemalloc to get the object allocation traceback
    
psycopg now acts the same way:
    
        $ python3 -Walways
        >>> import psycopg2
        >>> c = psycopg2.connect(database='psycopg2_test')
        >>> del c
        <stdin>:1: ResourceWarning: unclosed connection <connection object at 0x7fc4aec38d50; dsn: 'dbname=psycopg2_test', closed: 0>
        ResourceWarning: Enable tracemalloc to get the object allocation traceback
    
All warnings noticed during testing has been fixed.


dvarrazzo:
Hello, thank you.

I have some doubt about doing that for cursors. My doubts are:

- Normal cursors are not a heavy resources,
- but server-side cursors are,
- although closing the connection is all it takes to clear these resources.

I guess for you the case is that in case of doubt it's better to give a warning and let the user use a consistent closing pattern, right?

jdufresne:
> I guess for you the case is that in case of doubt it's better to give a warning and let the user use a consistent closing pattern, right?

For me, yeah.

Under normal circumstances, the ownership of all resources should be intentional, explicit, and deterministic. This warning helps identify the areas where that is not the case.

If there remains doubt, what do you think about me removing the cursor commit and emitting warnings for connections only? That would be easy to prepare.

We can always rethink warnings for cursors at a future time and I plan to use this PR for local testing anyway.

dvarrazzo:
I think the fact that in psycopg the _normal_ cursor happens to not be a resource holding more than its memory footprint is too much of a special case, and that switching from one type of cursor to the other shouldn't impact the program's pattern of interaction with psycopg. So it's probably fine to have the warning for these objects too.

jdufresne:
Sounds great!

I'll find some time to fix the Travis errors and the new cases discovered there.

psycopg/connection_type.c
```diff
@@ -126,10 +128,19 @@ psyco_conn_cursor(connectionObject *self, PyObject *args, PyObject *kwargs)
...
-    obj = NULL;
+    goto exit;
+
+error:
+    PyErr_Fetch(&error_type, &error_value, &error_traceback);
+    close = PyObject_CallMethod(obj, "close", NULL);
+    if (close)
+        Py_DECREF(close);
+    else
+        PyErr_WriteUnraisable(obj);
```
dvarrazzo: ISMT that if someone passes a factory that doesn't return a `cursor` object, the `IsInstance` test will fail, so we end up in this block, but the `obj.close()` could fail, because that's not a cursor, so the exception raised there would be clobbered.

maybe you should have two objects: `PyObject *obj = null` and `cursorObject *curs null`: once you check `isinstance(obj, cursor)` then you can set `curs = (cursorObject *)obj; obj = NULL;`, use the stronger type for the rest of the function, and use `Py_XDECREF` on both the functions on `exit:` (it's a pattern I prefer instead of tracing carefully the nullness of each var each possible branch of the function).

psycopg/connection_type.c
```diff
@@ -1366,6 +1377,9 @@ connection_dealloc(PyObject* obj)
 {
     connectionObject *self = (connectionObject *)obj;
 
+    if (PyObject_CallFinalizerFromDealloc(obj) < 0)
```
dvarrazzo: I don't know much about this method: I will try to read the docs later.

lib/_json.py
```diff
@@ -181,18 +181,18 @@ def _get_json_oids(conn_or_curs, name='json'):
...
-
-    # column typarray not available before PG 8.3
-    typarray = conn.info.server_version >= 80300 and "typarray" or "NULL"
-
-    # get the oid for the hstore
-    curs.execute(
-        "SELECT t.oid, %s FROM pg_type t WHERE t.typname = %%s;"
-        % typarray, (name,))
-    r = curs.fetchone()
+    with curs:
```
dvarrazzo: We don't necessarily want to close the cursor here. We expect the user passed us a cursor they want to keep playing with. Maybe the change should be made in `_solve_conn_curs()`

lib/_range.py
```diff
@@ -352,24 +352,24 @@ def _from_db(self, name, pyrange, conn_or_curs):
         from psycopg2.extensions import STATUS_IN_TRANSACTION
         from psycopg2.extras import _solve_conn_curs
         conn, curs = _solve_conn_curs(conn_or_curs)
+        with curs:
```
dvarrazzo: Ditto as in json

lib/extras.py
```diff
@@ -905,25 +905,25 @@ def get_oids(self, conn_or_curs):
         """Return the lists of OID of the hstore and hstore[] types.
         """
         conn, curs = _solve_conn_curs(conn_or_curs)
+        with curs:
```
dvarrazzo: same

psycopg/connection_type.c
```diff
@@ -60,6 +60,8 @@ psyco_conn_cursor(connectionObject *self, PyObject *args, PyObject *kwargs)
 {
     PyObject *obj = NULL;
     PyObject *rv = NULL;
+    PyObject *close;
```
dvarrazzo: I would prefer to use a `{block}` in the `error:` handler and the fine the variables in that scope.

lib/_json.py
jdufresne: I agree. I've opened #1040 to handle this. Let me know if this follows what you had in mind. If so, I'll rebase this PR.

jdufresne:
Thanks for the review. I believe I have addressed the feedback. All additional feedback is welcome.

I'm a bit confused as to why the tests are failing on PostgreSQL 8.1. IIUC, closing the cursor should be a noop. Do you have any ideas? I was unable to find a PostgreSQL 8.1 docker image for local testing, do you know of one?

dvarrazzo:
I don't know. The opportunity of developing new features on long unsupported database servers is not really a compelling one anyway.

However, yes, here there are ["packages" of old postgres versions for ubuntu](https://upload.psycopg.org/postgresql/): . The way to use them is just to untar them. See:

https://github.com/psycopg/psycopg2/blob/9bcca1a7b0e2cb793be0d4c7d98fdbaeba069128/scripts/travis_prepare.sh#L51-L58

jdufresne:
> The opportunity of developing new features on long unsupported database servers is not really a compelling one anyway

Are you open to dropping support for PostgreSQL server < 8.2? [It has been EOL since November 8, 2010](https://www.postgresql.org/support/versioning/). If you are, I'm happy to open a PR to make that happen. If not, I'll try to debug the issue further.

dvarrazzo:
> > The opportunity of developing new features on long unsupported database servers is not really a compelling one anyway
> 
> Are you open to dropping support for PostgreSQL server < 8.2? [It has been EOL since November 8, 2010](https://www.postgresql.org/support/versioning/). If you are, I'm happy to open a PR to make that happen. If not, I'll try to debug the issue further.

I think it's time to drop some old versions, yes, or at least stop to actively maintain the tests. But I'd like to look well into what range to maintain. Even if a postgres version is unsupported, it is not unheard of people not migrating them.

Maybe I'd try to have a chat in the mailing list.

About your failure: the problem is probably that the cursor is closed without any query performed, but the cursor is actually created on the server side at the first query. In those tests no operation is performed on the cursors. This is the relevant code.

https://github.com/psycopg/psycopg2/blob/9bcca1a7b0e2cb793be0d4c7d98fdbaeba069128/psycopg/cursor_type.c#L81-L103

In order to close the cursor only if exists we query a system table, but that table only exists from PG 8.2.

As far as I'm concerned you can either decorate those tests with `skip_before_postgres(8, 2)` or add a dummy "select 1" if the server is < 8.2 into the test to force the cursor to exist.

lib/_json.py
dvarrazzo: I am sorry, but here I was plain wrong. The cursor is one we are sure we have ownership here, whatever the input of `_solve_conn_curs()`.

As such, I think the changes is #1040 are not needed, and I wouldn't change the semantics of that function - maybe I'd just clarify it. I think the code is simpler using a `with curs` block here, where the cursor is used, rather than make `_solve_conn_curs()` return a context manager.

psycopg/connection_type.c
```diff
@@ -110,27 +110,44 @@ psyco_conn_cursor(connectionObject *self, PyObject *args, PyObject *kwargs)
...
         PyErr_SetString(PyExc_TypeError,
             "cursor factory must be subclass of psycopg2.extensions.cursor");
+        Py_DECREF(obj);
+        obj = NULL;
         goto exit;
     }
 
-    if (0 > curs_withhold_set((cursorObject *)obj, withhold)) {
-        goto exit;
+    curs = (cursorObject *)obj;
```
dvarrazzo: Here we have the same object assigned to both `curs` and `obj` so making sure we decref everything alright is tricky. I'm writing a patch for it.

jdufresne:
All tests now pass :tada: :cake: 

Thanks for all the help and reviews! I think it is ready for another round when you have the time.

dvarrazzo:
Thank you: will look into it :)

Issue #114, opened 2012-06-27
hstore adapter


psycoteam:
Originally submitted by: Psycopg website

Submitted by: Adrian Klaver

Seems the register_hstore only works with connections that use cursors that return an indexed tuple. When using the RealDictConnection/Cursor I get the traceback below. It can be worked around by supplying the OID.

```
/home/aklaver/<ipython-input-12-b3da0e399c73> in <module>()
----> 1 psycopg2.extras.register_hstore(cur)

/usr/local/lib/python2.6/dist-packages/psycopg2-2.4.3-py2.6-linux-i686.egg/
psycopg2/extras.pyc in register_hstore(conn_or_curs, globally, unicode, oid,
array_oid)
    742     """
    743     if oid is None:
--> 744         oid = HstoreAdapter.get_oids(conn_or_curs)
    745         if oid is None or not oid[0]:
    746             raise psycopg2.ProgrammingError(

/usr/local/lib/python2.6/dist-packages/psycopg2-2.4.3-py2.6-linux-i686.egg/
psycopg2/extras.pyc in get_oids(self, conn_or_curs)
    690 """ % typarray)
    691         for oids in curs:
--> 692             rv0.append(oids[0])
    693             rv1.append(oids[1])
    694
```

PR #1040, opened 2020-02-04
Improve resource ownership semantics of _solve_conn_curs()

jdufresne:
If a cursor is passed in, use it. Do not close it once finished.

If a connection is passed in, create a new cursor and close it once finished.


lib/extras.py
```diff
@@ -781,19 +782,26 @@ def wait_select(conn):
...
+@contextlib.contextmanager
 def _solve_conn_curs(conn_or_curs):
     """Return the connection and a DBAPI cursor from a connection or cursor."""
     if conn_or_curs is None:
         raise psycopg2.ProgrammingError("no connection or cursor provided")
 
     if hasattr(conn_or_curs, 'execute'):
         conn = conn_or_curs.connection
-        curs = conn.cursor(cursor_factory=_cursor)
+        curs = conn_or_curs
```
jdufresne: So it looks like this is intentionally always creating a new cursor and not using the passed in the cursor. Creating a new cursor is required for `test_non_dbapi_connection` to pass.

As a new cursor is always created, I think it should always be closed as well. Do you agree?

dvarrazzo: Yes, true. It works that way because sometimes user pass non-standard cursors, and if a non-DBAPI-compliant one is received the user function may fail (see #114, 21d323d2). The output of _solve_conn_curs() is now always a base class cursor.

dvarrazzo: And yes, of course a cursor we own can be safely closed (it won't be a server-side cursor, closing it won't cause any side effect).

lib/_json.py
```diff
@@ -180,23 +180,22 @@ def _get_json_oids(conn_or_curs, name='json'):
...
-    # get the oid for the hstore
-    curs.execute(
-        "SELECT t.oid, %s FROM pg_type t WHERE t.typname = %%s;"
-        % typarray, (name,))
-    r = curs.fetchone()
-
-    # revert the status of the connection as before the command
-    if conn_status != STATUS_IN_TRANSACTION and not conn.autocommit:
-        conn.rollback()
+    with _solve_conn_curs(conn_or_curs) as (conn, curs):
```
dvarrazzo: If you use this syntax, don't you also start a context on the connection, forcing a final commit/rollback?

jdufresne: No, it does not.  This enters the context manager `_solve_conn_curs()` which returns a tuple. This does not enter the context manager of the elements of the tuple. The context manager `_solve_conn_curs()`  also enters enters the cursor context manager, but not the one for the connection. Let me know if this is unclear and I can try to provide more details.

PR #1046, opened 2020-02-09
Remove support for PostgrSQL server < 8.2

jdufresne:


jdufresne:
Feel free to close this if the project isn't ready to take this step. Just thought I'd experiment with it a bit.

dvarrazzo:
As you had a run through the code to check for all the version comparisons, do you reckon 8.2 is the right version where to make the cut? Or there is good simplification to be had if we cut 8.3, 8.4 too etc?

psycopg/pqpath.c
```diff
@@ -360,8 +360,7 @@ pq_begin_locked(connectionObject *conn, PyThreadState **tstate)
     }
     else {
         snprintf(buf, bufsize,
-            conn->server_version >= 80000 ?
-                "BEGIN%s%s%s%s" : "BEGIN;SET TRANSACTION%s%s%s%s",
+                 "BEGIN%s%s%s%s",
```
dvarrazzo: I had this bit in mind more than any other.

Every other change is just about skipping tests, and about raising a cleaner error rather than an internal one because some function/construct doesn't exist.

But this is a different guy: if we drop this, psycopg will just stop working on these old servers (unless in autocommit).

I think I would leave the test here, because if someone is called to do some archeological work on some half-forgotten server deep in the heart of a temple in a jungle, then they could still use psycopg to operate on it, otherwise they'll also need to dig out an old psycopg version, or to do something else dangerous, which might result in a huge boulder rolling over them.

jdufresne:
I went with skipping the test on older servers. Thanks for you help figuring it out. I'll close this for now.

PR #1052, opened 2020-02-24
Converts page_size to int (closes #1051)

### LINKED ISSUE #1051: Allow page_size to use scientific notation
`psycopg2.extras._paginate()`  argument `page_size` must be an integer, so scientific notation cannot be used as they produce a float. This impacts `psycopg2.extras.execute_batch()` and `psycopg2.extras.execute_values()`.

```python
import psycopg2
from psycopg2.extras import execute_batch, execute_values

#... connect with cursor "cur"

# These work, as page_size is an integer
execute_values(cur,
    "insert into testfast (id, val) values %s",
    ((i, i * 10) for i in range(25)),
    page_size=10)
execute_values(cur,
    "insert into testfast (id, val) values %s",
    ((i, i * 10) for i in range(25)),
    page_size=10)

# These fail, as page_size isn't an integer
execute_values(cur,
    "insert into testfast (id, val) values %s",
    ((i, i * 10) for i in range(25)),
    page_size=1e1)
execute_values(cur,
    "insert into testfast (id, val) values %s",
    ((i, i * 10) for i in range(25)),
    page_size=1e1)
```

When page_size is a large integer, it's more convenient to use scientific notation than write out the whole integer. Would be great if psycopg2 just converted the input to integer.

amarvin:


amarvin:
#1051

PR #1060, opened 2020-03-11
Fix link to install docs in error message

tomasaschan:


dvarrazzo:
I don't understand? You didn't fix an error message, you changed the install file, and pointed it to a page that tells one lube and then points to the original page.

Rationale plz?

tomasaschan:
I got an error message that pointed me to a page that didn't exist; when I searched for the text in that error message it took me to this file, so I figured I'd update the link to point to something that does exist.

Now that I look more carefully at it, it seems that a) this text exists in multiple places, and I probably found the wrong one (I probably should have aimed for the one in the setup script...), and b) it has already been updated, since the link I tried to change here actually points to an existing page. I no longer have the error message in front of me, but my guess is that the version of the setup script was an older version, and that the URL has already been fixed (because all versions of this text I can find now have this URL, which is correct).

Sorry for the noise!

PR #1061, opened 2020-03-12
feat: Add AIX support

### LINKED ISSUE #105: Connection factories override cursor_factories
Originally submitted by: Corry Haines

The cursor subclasses appear to have been coded to override the cursor_factory kwarg that can be passed to them.

If this was the intended behavior, then I guess this is not a bug, but it seems to be a very odd behavior. Because of this behavior, adding a connection factory to an existing app caused most of the cursors to break when their request for specific cursors was ignored.

So, I have patched the subclasses to allow the cursor_factory kwarg to override the connection_factory. It allows connection factories to be dropped into place without stomping all over existing code.

I have uploaded the patch to my github clone of psycopg: https://github.com/tabletcorry/psycopg2/commit/eb46f4f80794e37a16e0848273ed12a0443ca2dd

abmusse:
AIX has the same issue as Solaris no .timeradd or .timersub

```
ld: 0711-317 ERROR: Undefined symbol: .timeradd

ld: 0711-317 ERROR: Undefined symbol: .timersub
```
This patch was derived from solaris_support.h and solaris_support.c

Resolves #105

### DIFF
--- a/psycopg/aix_support.c
+++ b/psycopg/aix_support.c
@@ -1,0 +1,58 @@
+/* aix_support.c - emulate functions missing on AIX
+ *
+ * Copyright (C) 2017 My Karlsson <mk@acc.umu.se>
+ * Copyright (c) 2018, Joyent, Inc.
+ * Copyright (C) 2020 The Psycopg Team
+ *
+ * This file is part of psycopg.
+ *
+ * psycopg2 is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In addition, as a special exception, the copyright holders give
+ * permission to link this program with the OpenSSL library (or with
+ * modified versions of OpenSSL that use the same license as OpenSSL),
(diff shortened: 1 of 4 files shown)


dvarrazzo:
I assume there isn't an equivalent for AIX of appveyor/travis right?

I can't test the patch myself so please make sure you run at least a `make check` before considering it completed, thank you

abmusse:
Here is the [test output](https://gist.github.com/abmusse/9eabf6463214b31637a3bf403250bb2b).

dvarrazzo:
Thank you @abmusse!

Can you figure out why some tests failed? Maybe the test database has an ASCII encoding? It doesn't seem an AIX-related problem though.

abmusse:
> Maybe the test database has an ASCII encoding? It doesn't seem an AIX-related problem though.

Yes I changed the enconding  to UTF-8.

Here is the updated [test output](https://gist.github.com/abmusse/4f72f8060baa61caa33a1b26cbf09308)

dvarrazzo:
Cool, thank you! I'll merge and release as soon as I can (which might be a few days because I'm out of town with limited visibility)

PR #1072, opened 2020-04-02
Add calling stored procedurs that have transaction control to list of…

draderaws:
Doc update - Include calling stored procedures that have transaction control inside to the list of SQL commands to be run with autocommit set to false.


dvarrazzo:
Can you add a link to the relevant bit of documentation in Postgres docs (not necessarily everywhere, at least from the most prominent place), and make sure you keep the text to < 80 cols? Thank you.

PR #1090, opened 2020-04-22
Added semver info, fixes #1053

### LINKED ISSUE #1053: Does this package follow semantic versioning?
Does this package follow [semantic versioning](https://semver.org) ([PEP 440](https://www.python.org/dev/peps/pep-0440/#semantic-versioning))? 
Is it safe to use ~= requirement operator?
```
psycopg2-binary~=2.8.4
```

alvassin:


dvarrazzo:
This is not needed. Semver is the default and sensible standard, we follow it without need to advertise their homepage in our front page.

PR #1101, opened 2020-05-04
Fix memory leak in conn_set_client_encoding

kmichel-sereema:
If the specified encoding is the same as the current one,
the early exit did not release the clean_enc string allocated
by clear_encoding_name.


dvarrazzo:
Good catch, thank you!

PR #1102, opened 2020-05-04
Find mxDateTime.h automatically

cthart:
We use mxDateTime in our old codebase. Right now the only way to compile this is by editing setup.cfg
This patch lets setup.py find the mxDateTime headers automatically.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,7 @@ What's new in psycopg 2.8.6
 
 - Fixed memory leak changing connection encoding to the current one
   (:ticket:`#1101`).
+- Fixed search of mxDateTime headers in virtualenvs (:ticket:`#996`).
 
 
 What's new in psycopg 2.8.5
--- a/setup.py
+++ b/setup.py
@@ -525,19 +525,21 @@ parser = configparser.ConfigParser()
 parser.read('setup.cfg')
 
 # check for mx package
-have_mxdatetime = False
 mxincludedir = ''
 if parser.has_option('build_ext', 'mx_include_dir'):
     mxincludedir = parser.get('build_ext', 'mx_include_dir')
 if not mxincludedir:
-    mxincludedir = os.path.join(get_python_inc(plat_specific=1), "mx")
+    # look for mxDateTime.h; prefer one located in venv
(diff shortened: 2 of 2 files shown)


dvarrazzo:
Thank you for this. So do you prefer automatic search over specifying a build option? If that's good for you it's ok for me.

I've squashed your changes and added a line into the news file. Once tests pass I'll merge it.

cthart:
Hi Daniele,

Sorry for my tardy response, I've been working on some other things the
last few days.

Thanks for applying my patch!

Well, you and I were both after the simplest way/patch that would work.
When I first started investigating this issue everything pointed to needing
to provide a parameter to pip when building. But the more I dug into it the
more apparent it became that the existing setup.py was simply not looking
in all possible places.

I tested with

pip install /home/colin/git/psycopg2

and then running

python -c "import psycopg2;
psycopg2.extensions.register_type(psycopg2._psycopg.MXDATETIME); print
'OK'";

(egenix-mx-base previously installed via pip of course)

So I'm confident this will also work when installed in the standard way
with "pip install psycopg"

When do you think the next release will be made? I'd like to start setting
up a new production server without any "hacks" to get it installed.

NB mxDateTime works with Python 3 so I don't expect it to go away anytime
soon, especially for people with large legacy codebases like ours. We also
have quite some services that use xmlrpc which pickles the data going
across, so you get the same data types on both sides. This makes it pretty
difficult to change to a regular datetime as then we would have to change
both sides of the service at the same time, potentially requiring us to
update as many as 20 locations... Oh well, fact of life, I guess :-)

Thanks again!

/Colin

dvarrazzo:
> I tested with
>
> pip install /home/colin/git/psycopg2 

I guess that's good enough. What I had in mind is:

    pip install -e git+https://github.com/psycopg/psycopg2.git#egg=psycopg2

but if you merged from upstream in your working copy that's good.

> NB mxDateTime works with Python 3 so I don't expect it to go away anytime
> soon, especially for people with large legacy codebases like ours.

Well, thank you for your input in suggesting how to make it more usable! I only used mx before pip existed: I didn't even had idea that the include files were installed via pip. 

I would like to make a new release in a couple of weeks.

Please consider sponsoring the project: see the sponsorship options on top of the page.

PR #1105, opened 2020-05-07
Handle failure in setup of ConnectionTwoPhaseTest

rafiss:
Previously, this test had a bug, because if the CREATE TABLE statement
failed, the setup would fail without committing or rolling back the
active transaction.

This is needed because `tearDown` is not called if there was an error
during `setUp` ([as specified by the `unittest` docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown)).

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -913,9 +913,11 @@ class ConnectionTwoPhaseTests(ConnectingTestCase):
             cur.execute("DROP TABLE test_tpc;")
         except psycopg2.ProgrammingError:
             cnn.rollback()
-        cur.execute("CREATE TABLE test_tpc (data text);")
-        cnn.commit()
-        cnn.close()
+        try:
+            cur.execute("CREATE TABLE test_tpc (data text);")
+            cnn.commit()
+        finally:
+            cnn.close()
 
     def count_xacts(self):
         """Return the number of prepared xacts currently in the test db."""


dvarrazzo:
Psycopg2 is a driver for postgres, which does support DDL in transaction. We don't plan to drop using that feature: writing tests supporting ancient versions of postgres is already painful enough.

dvarrazzo:
You are wrapping with blocks around the statements but you seem to think that those block will swallow the exception. That's not how it works: they will roll back the transaction and propagate the exception.

Can I kindly suggest you explain what you want to do? Thank you.

rafiss:
@dvarrazzo  Thank you for looking at my change! 

> Can I kindly suggest you explain what you want to do? Thank you.
I want to update this test so that if the `CREATE TABLE` statement fails, the transaction is rolled back appropriately.

> you seem to think that those block will swallow the exception. That's not how it works: they will roll back the transaction and propagate the exception.
I understand. I don't want to swallow any exceptions in my change. I was just interested in having the transaction rolled back if `CREATE TABLE` failed. Are you suggesting that I should keep the `try/except` logic to avoid having any exceptions bubble up?

Perhaps part of the confusion here was an oversight on my part. I had initially created a patch that used the `DROP TABLE IF EXISTS` syntax, which should never fail unless there is an exceptional condition. Then I realized this test must support older versions of Postgres that don't accept the `IF EXISTS` syntax.

> Psycopg2 is a driver for postgres, which does support DDL in transaction. We don't plan to drop using that feature: writing tests supporting ancient versions of postgres is already painful enough.
I understand. I don't want to make this change imply otherwise, but I didn't consider there would be a downside to using two transactions. I'm happy to keep using just one transaction in this test.

rafiss:
I see that because I got rid of the exception handling logic (and the `DROP IF EXISTS` syntax), that the tests all fail on the first run, because the initial `DROP` statement will always fail. 

I will create a new patch to account for this. Thanks for pointing this out to me.

dvarrazzo:
>  I was just interested in having the transaction rolled back if CREATE TABLE failed. 

My questions are:

- in what scenario CREATE TABLE would fail?
- if create TABLE TABLE fails, then the function `make_test_table()` didn't do its job, so the test cannot run. Why do you want to swallow the creation failure? What benefit has it got? If that table is not there the test wanting it will fail too.
- the connection is closed after the test is run, whether with success or failure. Why do you care about the transaction status of a connection that is bound to be closed?

dvarrazzo:
> if create TABLE TABLE fails, then the function `make_test_table()` didn't do its job, so the test cannot run. Why do you want to swallow the creation failure? What benefit has it got? If that table is not there the test wanting it will fail too.

[there you go](https://travis-ci.org/github/psycopg/psycopg2/jobs/684537532)
```
ERROR: test_tpc_rollback_recovered (tests.test_connection.ConnectionTwoPhaseTests)
----------------------------------------------------------------------
  File "tests/testutils.py", line 265, in skip_if_tpc_disabled_
    return f(self)
  File "tests/test_connection.py", line 1058, in test_tpc_rollback_recovered
    cur.execute("insert into test_tpc values ('test_tpc_commit_rec');")
UndefinedTable: relation "test_tpc" does not exist
LINE 1: insert into test_tpc values ('test_tpc_commit_rec');
                    ^
```

rafiss:
> My questions are:
> 
>     * in what scenario CREATE TABLE would fail?

To give some more background, I am doing this change because I work on [CockroachDB](https://www.cockroachlabs.com/docs/stable/postgresql-compatibility.html), which uses the PostgreSQL wire protocol. As such, Python users can use psycopg2 to work with CockroachDB. CockroachDB has a limitation in it such that one cannot drop and recreate a table in the same transaction. What was happening is the following:

- First test run creates the table successfully.
- Second test run fails at the `CREATE TABLE` step. The connection in `make_test_table` does not get closed, so the transaction remains open.
- Third test run reaches the `DROP TABLE` step, and blocks waiting for a lock on the table. The transaction from the second run is still open and holding the lock, so the test program is in deadlock.

However, that's just how I got here. It's possible the `CREATE TABLE` could fail for other reasons. In an extreme, unrealistic example, the limit on number of tables could be reached (if other queries are happening concurrently). It could also fail if the database runs out of storage space. Or the statement could succeed, but the client could timeout waiting for a response because of a temporary network failure.


>     * if create TABLE TABLE fails, then the function `make_test_table()` didn't do its job, so the test cannot run. Why do you want to swallow the creation failure? What benefit has it got? If that table is not there the test wanting it will fail too.

This is true. Swallowing the error doesn't seem like the right approach here. I didn't consider this because in my test runs, the `@skip_if_tpc_disabled` flag is `True`, so none of the tests actually get run. Although the `setUp` function _does_ still get run. It seems like this is because of how the `decorate_all_tests` function in testutils works. Perhaps that function be changed so it also skips test `setUp` and `tearDown` methods?

>     * the connection is closed after the test is run, whether with success or failure. Why do you care about the transaction status of a connection that is bound to be closed?

I don't believe that the connection is closed after a test run regardless of success or failure. If the `CREATE TABLE` fails, then the following two lines `cnn.commit()` and `cnn.close()` will not be executed. Thanks to you pointing this out, I now realize I don't need to concern myself with the transaction status. Closing the connection is what I care about. Perhaps the way I should be addressing this is by putting `cnn.close()` inside of a finally block.

---

I appreciate your point that psycopg2 is a driver for Postgres, and should not be concerned with adding additional functionality/refactoring to support other databases. However, I viewed this change as primarily making this test clean up after itself. The fact that it would make the test stop encountering a deadlock when running against CockroachDB is just an added benefit.

I hope some of my clarifications help! I really am very grateful that you have engaged with me on this change and have helped me realize a lot of edge cases and subtleties in this test that I did not consider before.

* Do you think you would accept this PR if I went with the approach is putting `cnn.close()` in a `finally` block? That pattern doesn't seem to be used elsewhere in the tests, so I'm not sure.
* I'm also curious if you have any thoughts on skipping `setUp` and `tearDown` if the test class is annotated as skipped. If you agree with doing that, I can create a separate PR for that.

dvarrazzo:
CREATE TABLE can definitely fail. What I mean is that it won't fail in normal conditions, and if it fails, the correct thing to do is to abort the test, not to soldier on. So the correct thing to do is to not catch the exception. In normal conditions CREATE TABLE doesn't fail. In abnormal conditions (disk full?) an exception is thrown and the test fails. That's it: we don't need to catch anything around it.

> I don't believe that the connection is closed after a test run regardless of success or failure. 

It gets closed on `tearDown()` of the `ConnectingTestCase` base class. `tearDown()` runs even if there was an exception in the test.

I don't mind if psycopg2 gets used with other protocol-compatible databases. However, as CockroachDB is a commercial software I expect at least some form of funding to the project to work for its support. Without it, I'm ok to give you the occasional help but not to disembowel the test suite to come forward the database differences, or to waste so much time on it.

rafiss:
> It gets closed on `tearDown()` of the `ConnectingTestCase` base class. `tearDown()` runs even if there was an exception in the test.

When using a debugger / logging, I observed that the `tearDown` function was not being called. I believe the reason is that `tearDown` is not called if there was an error during `setUp` ([as specified by the `unittest` docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown))

> I don't mind if psycopg2 gets used with other protocol-compatible databases. However, as CockroachDB is a commercial software I expect at least some form of funding to the project to work for its support. Without it, I'm ok to give you the occasional help but not to disembowel the test suite to come forward the database differences, or to waste so much time on it.

I hear your point about commercial software. In this change, I am planning to change anything about `psycopg2` functionality to support CockroachDB, nor am I planning to tremendously refactor the test suite. I don't want to waste your time either, but I really do appreciate the time you have given so far to help me out.

I am just hoping to make the connection that is created in `setUp` become closed if there is a failure during `setUp`. I'm happy to take whichever approach you recommend in order to achieve that. Would using a `try-finally` block in `setUp` be acceptable?

dvarrazzo:
> the reason is that tearDown is not called if there was an error during setUp

Ah fair enough.

so what you want is something like:

```python
with conn:
    # drop table
try:
    with conn:
        # create table
except Exception:
    self.tearDown()
    raise
```

perhaps?

I'm not against it: as long as the test suite keeps on passing for Postgres I'm fine. I'm not sure it solves your problem of CockroachDB (jeez, and I thought that PostgresSQL was an awful name).

> I hear your point about commercial software. In this change, I am not planning to change anything about psycopg2 functionality to support CockroachDB, 

I am not entirely ok of CockroachDB being able to brag compatibility with psycopg2 without contributing anything back. Let's say that merging your changes is an Enterprise Feature.

rafiss:
@dvarrazzo Do you have thoughts on this latest patchset I've created? It's just adding a finally block. (The reason being that `tearDown` is not called if there was an error in `setUp`.)

We had also discussed the possibility of modifying `testutils#decorate_all_tests` so that it skips `setUp` and `tearDown` if a class is annotated as skipped. This doesn't seem like a good approach, because some of the `skip_*` decorators rely on making a connection, and it is only possible to make a connection after `ConnectingTetstCase#setUp` is called.

dvarrazzo:
The last patch seems very little invasive. All the test grid passes so that sounds good on our part.

Is it everything you need? Ready to merge?

rafiss:
Yes, that's all I need. Thank you!

PR #1110, opened 2020-05-27
Fix ctypes doc example AttributeError

UpGado:
Original example ctypes code causes an error:
```python
libpq = ctypes.pydll.LoadLibrary(ctypes.util.find_library('pq'))
AttributeError: module 'ctypes' has no attribute 'util'
```
Adding `import ctypes.util` fixes the issue


dvarrazzo:
Thank you!

PR #1127, opened 2020-07-07
Handle failure in setup of IsolationLevelsTestCase

rafiss:
If the CREATE TABLE statement fails, the setup would fail
without committing or rolling back the active transaction, so the
transaction would hold onto its resources indefinitely.

Normally, the transaction would be closed when the connection is closed
in the `tearDown` function. However, `tearDown` is not called if there
was an error during `setUp` ([as specified by the `unittest` docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown)), so
we need to handle this case specially.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -567,9 +567,11 @@ class IsolationLevelsTestCase(ConnectingTestCase):
             cur.execute("drop table isolevel;")
         except psycopg2.ProgrammingError:
             conn.rollback()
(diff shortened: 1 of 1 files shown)


rafiss:
This is similar to the issue that was discussed here: https://github.com/psycopg/psycopg2/pull/1105

dvarrazzo:
All tests pass on regular postgres so that's good for me :+1:

PR #1131, opened 2020-07-26
use Py_SET_TYPE for compat with python 3.10

jouve:


psycopg/python.h
```diff
@@ -52,6 +52,14 @@ typedef long Py_hash_t;
 typedef unsigned long Py_uhash_t;
 #endif
 
+/* Since Py_TYPE() is changed to the inline static function,
+ * Py_TYPE(obj) = new_type must be replaced with Py_SET_TYPE(obj, new_type)
+ * https://docs.python.org/3.10/whatsnew/3.10.html#id2
+ */
+#if PY_VERSION_HEX < 0x030900A4
+  #define Py_SET_TYPE(obj, type) ((Py_TYPE(obj) = (type)), (void)0)
```
dvarrazzo, line 60: Thank you very much for the contribution!

I wonder why is the `(void)0` idiom preferred to the classic `do { ... } while (0)`: I have seen it accepted but not discussed [in the ticket](https://bugs.python.org/issue39573).

jouve: I think it makes the expression consistant with the definition of the inline function in 3.10 (it returns void):
```c
static inline void _Py_SET_TYPE(PyObject *ob, PyTypeObject *type) {
    ob->ob_type = type;
}
#define Py_SET_TYPE(ob, type) _Py_SET_TYPE(_PyObject_CAST(ob), type)
```

https://en.wikipedia.org/wiki/Comma_operator

jouve: but tbh, I reused the snippet as provided in python's release note ;)

PR #1136, opened 2020-08-04
Cockroachdb tests

dvarrazzo:


tests/test_async.py
```diff
@@ -350,6 +355,7 @@ def test_sync_poll(self):
         self.assertEquals(cur.fetchone()[0], 1)
 
     @slow
+    @skip_if_crdb
```
rafiss, line 358: Are the tests in this file missing a reason?

tests/testutils.py
```diff
@@ -407,6 +409,62 @@ def skip_if_windows(cls):
...
+            raise ValueError(
+                "can't parse CockroachDB version from %s" % sver)
+
+        ver = int(m.group(1)) * 10000 + int(m.group(2)) * 100 + int(m.group(3))
+        __crdb_version.append(ver)
+
+    return __crdb_version[0]
+
+
+def skip_if_crdb(reason, conn=None):
```
rafiss, line 441: I wonder if the decorator could also accept an optional issue number that refers to the CockroachDB repo. For most of the failing tests, we have an issue number that tracks why the test fails. See: https://github.com/cockroachdb/cockroach/blob/28c0337df4b695b6eba09639ca312a92348b9294/pkg/cmd/roachtest/psycopg_blocklist.go#L32

tests/testutils.py
```diff
@@ -407,6 +409,62 @@ def skip_if_windows(cls):
...
+    Can be used as a decorator for tests function or classes:
+
+        @skip_if_crdb("my reason")
+        class SomeUnitTest(UnitTest):
+            # ...
+
+    Or as a normal function if the *conn* argument is passed.
+    """
+    if conn is not None:
+        if crdb_version(conn) is not None:
```
rafiss, line 453: I wonder if we should also have a decorator that skips based on crdb_version. Something similar to `skip_before_postgres`

tests/test_async.py
dvarrazzo: I'll double check. I assume because `notify` is not supported

tests/testutils.py
dvarrazzo: We can surely make it so. I was waiting to have a use case for it. Do you have it? Is there a feature that you know unsupported in certain CRDB versions and supported in others, which I could test using a docker image?

tests/testutils.py
dvarrazzo: Mmm sounds good, I can add it and use the go file to populate them

tests/test_async.py
dvarrazzo: That was the case, however the reason was missing. Added missing reasons, and fixed decorator in order to require one.

tests/testutils.py
dvarrazzo: Added. My list of bugs is slightly different: many were marked failed e.g. because temp table weren't supported: I think that feature was added later. I've compiled a hopefully more up-to-date list.
```
     28 not supported on CockroachDB: 2-phase commit (https://github.com/cockroachdb/cockroach/issues/22329)
      4 not supported on CockroachDB: backend pid (https://github.com/cockroachdb/cockroach/issues/35897)
      3 not supported on CockroachDB: cancel (https://github.com/cockroachdb/cockroach/issues/41335)
      1 not supported on CockroachDB: cast adds tz (https://github.com/cockroachdb/cockroach/issues/51692)
      3 not supported on CockroachDB: cidr (https://github.com/cockroachdb/cockroach/issues/18846)
     13 not supported on CockroachDB: composite (https://github.com/cockroachdb/cockroach/issues/27792)
      3 not supported on CockroachDB: connect any db
     19 not supported on CockroachDB: copy (https://github.com/cockroachdb/cockroach/issues/41608)
      1 not supported on CockroachDB: deferrable (https://github.com/cockroachdb/cockroach/issues/48307)
      1 not supported on CockroachDB: diagnostic
      3 not supported on CockroachDB: encoding (https://github.com/cockroachdb/cockroach/issues/35882)
      2 not supported on CockroachDB: flush on write flakey
      9 not supported on CockroachDB: hstore (https://github.com/cockroachdb/cockroach/issues/41284)
      1 not supported on CockroachDB: infinity date (https://github.com/cockroachdb/cockroach/issues/41564)
      1 not supported on CockroachDB: interval style (https://github.com/cockroachdb/cockroach/issues/35807)
     25 not supported on CockroachDB: isolation level
     16 not supported on CockroachDB: json (https://github.com/cockroachdb/cockroach/issues/)
     45 not supported on CockroachDB: large objects (https://github.com/cockroachdb/cockroach/issues/243)
     40 not supported on CockroachDB: named cursor (https://github.com/cockroachdb/cockroach/issues/41412)
      5 not supported on CockroachDB: nested array (https://github.com/cockroachdb/cockroach/issues/32552)
      8 not supported on CockroachDB: notice
     14 not supported on CockroachDB: notify (https://github.com/cockroachdb/cockroach/issues/41522)
     14 not supported on CockroachDB: range (https://github.com/cockroachdb/cockroach/issues/41282)
      1 not supported on CockroachDB: stored procedure (https://github.com/cockroachdb/cockroach/issues/1751)
      1 not supported on CockroachDB: table oid
```

tests/testutils.py
dvarrazzo: Now it's possible to specify when to skip a test using something like:
```
    @skip_if_crdb("notify", version="<= 20.2")
```
Not used anywhere yet: do you have any place where to use it already?

rafiss: Thanks! One place where it could be used is for enums, which we will only begin supporting in v20.2. However, I don't know if any tests rely on that specifically.

PR #1165, opened 2020-10-12
Remove unnecessary sensationalism in docs

hackedhead:
SQL injection and failures are bad but this seems over the top. In a
literal sense I'm struggling to imagine a scenario where bad
interpolation is worth getting shot over. And I understand this was
probably intended as hyperbole, but I don't think there's any need to
invoke the spectre of violence to make the point here.

### DIFF
--- a/doc/src/usage.rst
+++ b/doc/src/usage.rst
@@ -211,7 +211,7 @@ reliable. We must stress this point:
 
     Never, **never**, **NEVER** use Python string concatenation (``+``) or
     string parameters interpolation (``%``) to pass variables to a SQL query
-    string.  Not even at gunpoint.
+    string.
 
 The correct way to pass variables in a SQL command is using the second
 argument of the `~cursor.execute()` method::


dvarrazzo:
Can you please replace it with a different hyperbole?

hackedhead:
I agree with the sentiment of the warning, and I commonly _use_ hyperbole to reinforce my ideas as well, but only in settings where I fundamentally know my audience personally, because satire is a complex language construction, and it can be easily misconstrued in text communication.

For this reason, I think documentation (and especially a warning) is not really an appropriate place for hyperbole generally. The rest of the documentation on SQL parameter usage seems very good, and it seems a shame to confuse it with satire. The page is already (earnestly) using the red warning box, font cues and repetition for emphasis against concatenation and interpolation, with good explanations of the problems they lead to. My feeling is that the presence of hyperbole is is more likely to confuse, rather than aid, the understanding of the uninitiated via [Poe's Law](https://en.wikipedia.org/wiki/Poe%27s_law). Since the uninitiated are the primary target of this warning in general, it seems a mistake to effectively have an "in-joke" in the warning.

dvarrazzo:
@hackedhead,

thank you very much for your interest. However, as you correctly realised, that one is an hyperbole. I expect users of the adapter to be able, as you are, to understand it. If they are not, it is frankly not a problem that can be effectively solved by a database adapter.

That red box has been even [twitted around](https://twitter.com/search?q=psycopg%20gunpoint&src=typed_query) and laughed about: it seems to me that it has fulfilled its role: people may remember about it, when they write their query and avoid an SQL injection bug.

I would rather not get too deep into a personal ramble about the cultural landscape where such conversation has a need to be brought up; I will limit to point out that it is not a widely shared opinion that such joke, as poor as it is, is inappropriate. As you brought up the topic, I thought about whether I really find it inappropriate and, as much as I am sympathetic with many of the battles that fellow human beings are fighting for, I don't believe that this database adapter's documentation should have a lame joke censored because someone in a country with a poor gun control policy might hypothetically be offended.

As such, I prefer to leave that line it where it is. I am available to replace with a more subtle but equally memorable line, if you have one to suggest.

PR #1171, opened 2020-10-19
Docs: Fix syntax error in code example

jvamvas:


asqui:
Why is this a syntax error?

The semicolon is superfluous both at the end of the query, and also at the end of the line of Python code, so it might be best to remove it altogether. (Unless I'm missing something?)

dvarrazzo:
It should be removed altogether :)

dvarrazzo:
Cherry-picked, thank you.

PR #1179, opened 2020-10-30
Ignore decode errors in conn_decode()

dlax:
The previous error handling mode (the default "strict" mode) is problematic when trying to access data with different encoding, for instance from the `pg_stat_activity` view as shown in the following issue in pg_activity: https://github.com/dalibo/pg_activity/issues/149#issuecomment-718476674.

With this change, decoding errors are now ignored, so the behavior is similar to psql.

### DIFF
--- a/psycopg/connection_int.c
+++ b/psycopg/connection_int.c
@@ -126,15 +126,17 @@ conn_decode(connectionObject *self, const char *str, Py_ssize_t len)
 
     if (self) {
         if (self->cdecoder) {
-            return self->cdecoder(str, len, NULL);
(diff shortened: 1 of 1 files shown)


dvarrazzo:
That's not going to happen, no.

Use the `BYTES` typecaster to avoid the decoding, so that you will receive your text data in bytes, and then you can do the decoding yourself and ignore errors if you wish.

```
psycopg2.extensions.register_type(psycopg2.extensions.BYTES, context)  # context can be a connection or cursor
```

see https://www.psycopg.org/docs/faq.html#faq-bytes.

PR #1188, opened 2020-11-11
Fix typo in sample code in install.rst: psycopg >> psycopg2

ic4f:
I believe that's a typo?


dvarrazzo:
I guess it is, thank you

PR #1190, opened 2020-11-12
Add support for PostgreSQL 13 (final)

martinfrancois:
Since 2020-09-24 PostgreSQL is no longer the future 😃 

I added PostgreSQL 13 to run as part of the build pipeline and updated the docs to reflect support for version 13 as well.

Feel free to tell me if anything is missing, thanks!

### DIFF
--- a/scripts/travis_prepare.sh
+++ b/scripts/travis_prepare.sh
@@ -125,6 +125,7 @@ if (( "$TEST_PRESENT" )); then
     # Postgres versions supported by ARM64
         create 10
     else
+        create 13
         create 12
         create 11
         create 10
@@ -150,5 +151,6 @@ fi
 
 # Postgres built from master
 if (( "$TEST_FUTURE" )); then
-    create 13 13-master
+    # create 14 14-master
+    true
 fi
--- a/scripts/travis_test.sh
+++ b/scripts/travis_test.sh
@@ -56,6 +56,7 @@ if (( "$TEST_PRESENT" )); then
         run_test 10
     else
     # Postgres versions supported by Travis CI
+        run_test 13
         run_test 12
         run_test 11
         run_test 10
@@ -81,5 +82,6 @@ fi
 
 # Postgres built from master
 if (( "$TEST_FUTURE" )); then
-    run_test 13
+    # create 14 14-master
+    true
 fi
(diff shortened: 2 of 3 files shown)


scripts/travis_prepare.sh
```diff
@@ -147,8 +148,3 @@ if (( "$TEST_PAST" )); then
     create 9.2
     create 9.3
 fi
-
-# Postgres built from master
-if (( "$TEST_FUTURE" )); then
```
dvarrazzo, line 152: Can you please leave this stanza here to remember how to add back 14? :)

dvarrazzo (changes requested):

dvarrazzo:
Is PG13 already available on travis? Did you run a build?

martinfrancois:
> Is PG13 already available on travis? Did you run a build?

Yes I did, and yes it is! :)

scripts/travis_prepare.sh
martinfrancois: Sure! :)

martinfrancois:
@dvarrazzo I just updated it with your change requests, did I do it as you meant? :)

scripts/travis_prepare.sh
martinfrancois: Done!

dvarrazzo:
Not really :) I meant to leave the if there, but empty. Now, bash seems it doesn't really allow an empty if, so I think it should be like:

```
# Postgres built from master
if (( "$TEST_FUTURE" )); then
    # create 14 14-master
    true
fi
```

so that we know what to do as soon as 14 starts blipping on the radar.

martinfrancois:
> # Postgres built from master
> if (( "$TEST_FUTURE" )); then
>     # create 14 14-master
>     true
> fi
@dvarrazzo alright, got you now! What do you think about it now?

scripts/travis_test.sh
```diff
@@ -81,5 +82,6 @@ fi
 
 # Postgres built from master
 if (( "$TEST_FUTURE" )); then
-    run_test 13
+    # create 14 14-master
```
dvarrazzo, line 85: This is the wrong comment, but nevermind, it's good ok.

martinfrancois:
@dvarrazzo thanks for the quick handling and merge of the PR! 👍

PR #1192, opened 2020-11-16
Removed Windows builds for Python 2.7/3.4 and link against libpq 13.0

jerickso:
Removed support for building Python 2.7 and 3.4 binaries in AppVeyor and changed PostgreSQL libpq library linkage from version 11.x to 13.0.  Version of Postgres > 11 do not have build support for Visual Studio versions before 2013 requiring dropping support of versions of Python that are built against older version (Python 2.7 uses 2008 and Python 3.4 uses 2010).


dvarrazzo:
Looks good, all tests pass. Thank you very much!

PR #1193, opened 2020-11-16
Allow Windows building against the lastest Postgres 11.x code

jerickso:
Changes to allow building the Window binaries in AppVeyor against the latest version of the Postgres 11.x code, currently at version 11.10.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -27,7 +27,7 @@ environment:
         - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015, PY_VER: "27", PY_ARCH: "64"}
 
     OPENSSL_VERSION: "1_1_1h"
-    POSTGRES_VERSION: "11_4"
+    POSTGRES_VERSION: "11_10"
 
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Hi @jerickso if we merged this branch it would mean that we wouldn't be able to build wheel packages of the next 2.8.x for Python 2.7, right? If that's the case then it's better to limit the changes you made to the windows build to master, and only release them in 2.9.

jerickso:
No, I expect these patches will still be able to build wheel packages for Python 2.7 as long as it is built against Postgres 11.X in the wheels repository.  I can understand the reasoning behind the question as the majority of changes are the same as the pull request I submitted for the master branch with the only difference being the version of Postgres that we build/link against.  It will break for Python 2.7 if we try to build against Postgres 12.x or later.

Here's my AppVeyor build results in psycopg2 of the changes I submitted:
[https://ci.appveyor.com/project/jerickso/psycopg2/builds/36303063](https://ci.appveyor.com/project/jerickso/psycopg2/builds/36303063)

If you want, I can verify these changes against the wheels repository as well.

Issue #1196, opened 2020-11-17
Drop Python 3.5 EOL


auvipy:

hugovk:
Please see PR https://github.com/psycopg/psycopg2/pull/1197.

PR #1197, opened 2020-11-17
Drop support for EOL Python 3.5

hugovk:
Fixes https://github.com/psycopg/psycopg2/issues/1196.


psycopg/python.h
```diff
@@ -34,8 +34,8 @@
 #error "psycopg requires Python 2.7"
 #endif
 
-#if PY_3 && PY_VERSION_HEX < 0x03050000
-#error "psycopg requires Python 3.5"
+#if PY_3 && PY_VERSION_HEX < 0x03060000
```
dvarrazzo, line 37: Unrelated, but in this file there is an ifdef to Python 3.2: could you please drop it?

dvarrazzo (changes requested):
Looks good. There's sure tons of stuff we will be able to get rid of by dropping 2.7 instead :)

dvarrazzo:
Ah, I didn't notice it earlier, but in the NEWS file could you format the title with `---` instead of `^^^` (which will give it the right level in the documentation), and call it 3.9 instead of 3.9.0? See release 3.8 for comparison. Thank you

psycopg/python.h
hugovk: This bit? Isn't it used by Python 2.7?


```h
/* hash() return size changed around version 3.2a4 on 64bit platforms.  Before
 *   this, the return size was always a long, regardless of arch.  ~3.2
 *   introduced the Py_hash_t & Py_uhash_t typedefs with the resulting sizes
 *   based upon arch. */
#if PY_VERSION_HEX < 0x030200A4
typedef long Py_hash_t;
typedef unsigned long Py_uhash_t;
#endif
```

dvarrazzo: Ah right, I guess it is. I thought it was into an "if PY3" branch.

dvarrazzo:
That's gone too! Are you planning about butchering 2.7 too? That's going to be fun...

hugovk:
Yep, already started on it!

PR #1198, opened 2020-11-17
Drop support for EOL Python 2.7

hugovk:
Also used https://github.com/asottile/pyupgrade with `--py36-plus` to upgrade Python syntax.


doc/src/tools/make_sqlstate_docs.py
```diff
@@ -40,7 +39,7 @@ def main():
     for l in lines:
         cls = l.sqlstate[:2] if l.sqlstate else None
         if cls and cls != sqlclass:
-            print("**Class %s**: %s" % (cls, sqlclasses[cls]))
+            print("**Class {}**: {}".format(cls, sqlclasses[cls]))
```
dvarrazzo, line 42: Couldn't use f-strings instead of format?

hugovk: Yes, could do. I used https://github.com/asottile/pyupgrade:

> _note_: `pyupgrade` is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are anything but simple names or dotted names (as this can decrease readability).

I've just pushed a new commit that upgraded more using https://github.com/ikamensh/flynt.

Flynt is also cautious in places -- https://github.com/ikamensh/flynt#dangers-of-conversion -- but has an option to run for more. I've put those in a commit in another branch: https://github.com/hugovk/psycopg2/commit/f8d20e1fe4946c7bb6ffa7b55c8e0a735b634b00. What do you think of those?

dvarrazzo: In my opinion `str.format()` has not practical reason to exist, whereas I regard f-strings the best invention bar sliced bread and maybe a couple of others, but that's only personal. So for me every % operator should either become an f-string or stay as it is.

Your last changeset could do without the `int()` where the format doesn't force explicitly the number of chars with a format: for instance the value:

            return msg + _os.linesep + "  (execution time: %d ms)" % t

is converted to:

            return msg + _os.linesep + f"  (execution time: {int(t)} ms)"

but it would be enough:

            return msg + _os.linesep + f"  (execution time: {t} ms)"

although if the real deal would be:

            return f"{msg}{os.linesep}  (execution time: {t} ms)"

Your changeset seems useful to find all the occurrences though, and they are not many, about 15. Would you take up the chore of make them human? I can do it on top of your changeset if not.

hugovk: Updated! Let me know if there's more you'd like changed, and feel free to work on top of this as well :)

Issue #829, opened 2018-12-21
psycopg2.OperationalError: server closed the connection unexpectedly


RekGRpth:
this code

```python
sql = 'select pg_backend_pid()'

class DAL():
    def __init__(self, uri):
        from psycopg2 import connect
        #from pg8000 import connect
        self.cursor = connect(user='postgres', host='postgres').cursor()

    def executesql(self, sql):
        self.cursor.execute(sql)
        return self.cursor.fetchall()

def thread():
    db = DAL('')
    i = 10
    import time
    from pprint import pprint
    while i > 0:
        pprint(dict(function='thread:run', executesql='%s' % db.executesql(sql)[0][0]))
        time.sleep(1)
        i -= 1

def process():
    from pprint import pprint
    pprint(dict(function='process:begin'))
    import time
    time.sleep(2)
    pprint(dict(function='process:end'))

def main():
    from pprint import pprint
    pprint(dict(function='main:begin'))
    import threading
    threading.Thread(target=thread, name='mythread').start()
    import time
    time.sleep(2)
    import multiprocessing
    multiprocessing.Process(target=process, name='myprocess').start()
    pprint(dict(function='main:end'))

if __name__ == '__main__': main()
```

raises following exception

```python
Exception in thread mythread:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "test3.py", line 19, in thread
    pprint(dict(function='thread:run', executesql='%s' % db.executesql(sql)[0][0]))
  File "test3.py", line 10, in executesql
    self.cursor.execute(sql)
psycopg2.OperationalError: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

but with pg8000 everething is ok

RekGRpth:
also, with python2 everething is ok

dvarrazzo:
try keeping the connection alive. instead of:

        self.cursor = connect(user='postgres', host='postgres').cursor()

try:

        self.connection = connect(user='postgres', host='postgres')
        self.cursor = self.connection.cursor()

RekGRpth:
result is same

dimandzhi:
[...]

PR #1205, opened 2020-12-05
Handle forking with connection pools more safely

jmo-qap:
https://www.psycopg.org/docs/usage.html#thread-and-process-safety reminds you that you shouldn't use libpq connections made in the parent in a fork. However, sometimes one doesn't control the forking and then one spends a lot of time debugging ... :-| This PR makes a connection pool to help catch a situation where a pre-fork connection pool is used post-fork and generally avoid re-using connections.

This is a Request for Comment and the limitation is that it is only for connection pools and doesn't handle the case where you've already acquired a connection from the pool pre-fork. I think there are other ways to detect this event but figured this was a start of an approach.

### DIFF
--- a/lib/pool.py
+++ b/lib/pool.py
@@ -185,3 +185,24 @@ class ThreadedConnectionPool(AbstractConnectionPool):
             self._closeall()
         finally:
             self._lock.release()
+
+class MultiProcessThreadedConnectionPool(ThreadedConnectionPool):
+
+    def __init__(self, *args, **kwargs):
+        self.pid = os.getpid()
+        self.args = args
+        self.kwargs = kwargs
+        super().__init__(*args, **kwargs)
+
+    def getconn(self, *args, **kwargs):
+        current_pid = os.getpid()
+        if current_pid != self.pid:
+            super().__init__(*self.args, **self.kwargs)
+            self.pid = current_pid
+        return super().getconn(*args, **kwargs)
+
+    def putconn(self, *args, **kwargs):
+        current_pid = os.getpid()
+        if current_pid != self.pid:
+            raise Exception("Returned a connection to the pool that was acquired pre-fork. Don't use pre-fork connections post-fork.")
+        return super().putconn(*args, **kwargs)


lib/pool.py
```diff
@@ -185,3 +185,24 @@ def closeall(self):
...
+    def __init__(self, *args, **kwargs):
+        self.pid = os.getpid()
+        self.args = args
+        self.kwargs = kwargs
+        super().__init__(*args, **kwargs)
+
+    def getconn(self, *args, **kwargs):
+        current_pid = os.getpid()
+        if current_pid != self.pid:
+            super().__init__(*self.args, **self.kwargs)
```
dvarrazzo, line 200: It's not good to just call `__init__` on a class already initialised. I think if you detect this state you should empty the pool: you could add a protected `_flush_pool()` method on the base class to encapsulate it.

lib/pool.py
```diff
@@ -185,3 +185,24 @@ def closeall(self):
             self._closeall()
         finally:
             self._lock.release()
+
+class MultiProcessThreadedConnectionPool(ThreadedConnectionPool):
```
dvarrazzo, line 189: Nobody would use a specific subclass if they haven't already gone through some painful debug process as you say. So maybe this should be a change to the base class, not the addition of a new class.

lib/pool.py
```diff
@@ -185,3 +185,24 @@ def closeall(self):
...
+        current_pid = os.getpid()
+        if current_pid != self.pid:
+            super().__init__(*self.args, **self.kwargs)
+            self.pid = current_pid
+        return super().getconn(*args, **kwargs)
+
+    def putconn(self, *args, **kwargs):
+        current_pid = os.getpid()
+        if current_pid != self.pid:
+            raise Exception("Returned a connection to the pool that was acquired pre-fork. Don't use pre-fork connections post-fork.")
```
dvarrazzo, line 207: This should probably be a `ProgrammingError`

dvarrazzo (changes requested):

dvarrazzo:
I wonder if an anti-fork protection shouldn't be baked into the connection itself. There is already something partial: forking is detected on delete, see #829, so the connection, internally, has the pid it was created on. If we expose that to python, would you be able to do more robust checking?

dvarrazzo:
Closing as this is a connection problem, not a pool problem.

jmo-qap:
👍  Will circle back on this when I get a chance to check on connection status. I think a connection-based approach would just check if the init pid is different than the current PID and raise an error. Wouldn't want to create a new connection since that would potentially introduce confusing semantics about previous connection state.

PR #1206, opened 2020-12-05
Show how to troubleshoot common issues

mtangoo:
While building the documentation I met some errors and warnings that were not documented and only found out when dug  a little bit. Am here by making them more visible to make it convenient for new users

### DIFF
--- a/doc/README.rst
+++ b/doc/README.rst
@@ -18,3 +18,23 @@ Then you can build the documentation with::
     make
 
 You should find the rendered documentation in the ``html`` directory.
+
+Troubleshooting
+----------------------------------
+1. Outdated pip version
+If you get warning of having old pip version be sure to upgrade it.
+>WARNING: You are using pip version 20.2.4; however, version 20.3.1 is available.
+>You should consider upgrading via the 'PATH_TO_ROOT_FOLDER/psycopg2-2.8.6/doc/env/bin/python -m pip install --upgrade pip' command.
+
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Were you trying to build the docs with Python 2? That's probably no more supported

dvarrazzo:
Wouldn't it be better to make the Makefile work than to write in the readme all the possible ways it can fail?

dvarrazzo:
...for instance, regarding the pth, maybe it's possible to add something like `-e ../setup.py` in the requirements to have it installed in development mode more robustly?

mtangoo:
Hi @dvarrazzo, sorry for late reply. I was mostly offline

>Were you trying to build the docs with Python 2? That's probably no more supported

No! I had Python3.9 and Python 2.7 where 2.7 is the default wrecking all the havoc. I haven't looked at Python with any database save SQLite for years so I was perplexed. Then digging this is what I found.

>Wouldn't it be better to make the Makefile work than to write in the readme all the possible ways it can fail?

Two reasons (correct me if am wrong). First I assumed there are few limited way the compilation can go wrong (if someone follows instructions correctly) mostly related to Python version conflicts. Two I thought that rather than trying to support all Python versions on Make file, setting ENV for example is easier and flexible

>...for instance, regarding the pth, maybe it's possible to add something like -e ../setup.py in the requirements to have it installed in development mode more robustly?

If any of those can be resolved by updating README, then it is fine. I just didn't bother to look further into it

dvarrazzo:
Going about to merge old requests. Building the docs is not something I would like to offer a public interface for and I don't think that confusing Python version is the only thing that can go wrong. So I prefer to avoid merging this. Thank you very much.

PR #1247, opened 2021-02-28
ci: first stab at running CI as github actions

xrmx:
There are some differences between this and travis configuration:
- Python 2.7 is treated as the others, can't understand travis config aim
- There's no arm64 workers in github actions
- Cannot test docs building because you have to set the token
- Notifications options documented here https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/configuring-notifications#github-actions-notification-options

Stuff can be simplified a bit more later when it's not needed to keep travis running.

Props to @AdamChainz for sharing its setup.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,17 @@
+name: Build documentation
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  docs:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - name: install requirements
+      run: pip install --upgrade pip setuptools wheel
+    - name: run tests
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Hello! Ciao Riccardo, Thank you for this. 

I assume you have read the psycopg tweet. What I meant was more thinking about psycopg3, because at the moment I'm actively developing on it. Target is only Python 3.6-3.9 there (Python 2.7 should be dropped on psycopg2 master too anyway).

Would you be interested in putting together GitHub Actions for psycopg3, then we can review psycopg2 later?

dvarrazzo:
Hello! I have recovered this branch, integrated the changes we made for psycopg3 and pushed it to the maintenance branches.

https://github.com/psycopg/psycopg2/compare/bbe42270...56cd4709

Thank you very much again!

PR #1255, opened 2021-04-04
Fixes #1252 - Increment refcount on provided parameters.

### LINKED ISSUE #1252: Creating `psycopg2.extensions.Column` from Python leads to a segfault
While working on some test code that creates a `psycopg2.extensions.Column` object from Python code we discovered that creating this object is enough to trigger a segfault when the object is deleted. Admittedly the library only creates that object from C code when building the result of `cursor.description()` so the use case isn't entirely realistic.

Here's an example that demonstrates the crash. I've confirmed it behaves the same on Python 3.8 from Ubuntu 20.04 and Python 3.9 from Fedora 34.
```python
Python 3.9.2 (default, Feb 20 2021, 00:00:00) 
[GCC 11.0.0 20210210 (Red Hat 11.0.0-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from psycopg2.extensions import Column
>>> def do_stuff():
...     foo = Column(name='my_column')
... 
>>> for _ in range(1000):
...     do_stuff()
... 
Segmentation fault (core dumped)
```

Here's the virtual environment for running the test:
```
$ pip3 freeze
psycopg2==2.8.6
```

byronclark:
When parsing an object (`O`) with `PyArg_ParseTuple` and friends, the
provided object is set in the appropiate field, but the refcount is not
incremented. Reference:
https://docs.python.org/3/c-api/arg.html#other-objects.

Use `Py_XINCREF` to increment the refcount on any objects that may have
been passed into this function. `Py_XINCREF` performs a NULL check
internally and only increments the refcounter on non-NULL objects.

### DIFF
--- a/psycopg/column_type.c
+++ b/psycopg/column_type.c
@@ -108,6 +108,16 @@ column_init(columnObject *self, PyObject *args, PyObject *kwargs)
         return -1;
     }
 
+    Py_XINCREF(self->name);
+    Py_XINCREF(self->type_code);
+    Py_XINCREF(self->display_size);
+    Py_XINCREF(self->internal_size);
+    Py_XINCREF(self->precision);
+    Py_XINCREF(self->scale);
+    Py_XINCREF(self->null_ok);
+    Py_XINCREF(self->table_oid);
+    Py_XINCREF(self->table_column);
+
     return 0;
 }
 


dvarrazzo:
Fixed in 8ea5d0c0b8e7dc74e516ae24c3841a7c4b5a0d21

byronclark:
Thank you for the fix!

PR #1278, opened 2021-05-22
Fix typo: catched -> caught

yhay81:


dvarrazzo:
Why closed?

yhay81:
Sorry for disturbing. I googled the word "catched" and found some cases, which says "error is catched" is not unusual.

dvarrazzo:
Yes, caught is a better wording. Please reopen and I'll merge it :)

PR #1292, opened 2021-06-16
Add s back to placeholder string

plally:
execute and mogrify error due to an s not being included at the end of named placeholder strings.
in prior versions sql.Placeholder included an `s` at the end of named placeholder strings. it no longer does this. 
I am making this pr under the assumption that this was done in error.


dvarrazzo:
Fixing this on my side, adding a test too.

PR #1315, opened 2021-07-01
Try to switch pkg-config (works on OpenSUSE)

frknkrc44:


setup.py
```diff
@@ -102,10 +107,48 @@ def __init__(self, build_ext):
...
+        elif attr_name == "includedir":
+            process = "{} --cflags libpq | sed 's/^-I//'".format(self.pkg_config_exe)
+        elif attr_name == "includedir-server":
+            process = "{0} --cflags libpq | sed -e 's/^-I//' -e 's/\s$/\/'$({0} --modversion libpq | cut -f1 -d\".\")'\/server/'".format(self.pkg_config_exe)
+        elif attr_name == "version":
+            process = "{} --modversion libpq | sed 's/^/PostgreSQL /'".format(self.pkg_config_exe)
+        elif attr_name == "ldflags":
+            process = "{} --libs-only-L libpq".format(self.pkg_config_exe)
+        elif attr_name == "cppflags":
+            process = "{} --cflags libpq".format(self.pkg_config_exe)
```
dvarrazzo, line 129: I am sorry but this is really not something that I wish to accept. Depending on so many external tool is nothing I wish to maintain. The pg_config exists exactly to avoid this.

If you provide a patch so that, calling `pg_config --what` it looks for an env var `PG_CONFIG_WHAT` and uses it in preference then we might consider it. Then Suse or whatever else think it's a good idea could create an external wrapper using cut, sed, awk, perl, or whatever bashism to call `setup.py`.

dvarrazzo:
Closing as the proposed approach is not portable nor maintainable.

PR #1339, opened 2021-08-19
Feature/context managers

FelipeV34:


dvarrazzo:
You should explain what you are trying to achieve

FelipeV34:
I added context managers to manage the connection and cursor resources automatically.

Allows you to write code like 

```
with connection(**db_credentials) as conn:
    with cursor(conn) as cur:
        # db calls using the cursor
```

Or like

```
with connection_cursor(**db_credentials) as cur:
    # db calls using the cursor
```

And it clases the resources at the end of the block.

dvarrazzo:
The cursor context manager is redundant: cursors already work that way.

I would prefer to not add these features to psycopg2. Psycopg3 has a set of better designed context managers and I would rather people to use them.

https://www.psycopg.org/psycopg3/docs/basic/usage.html#main-objects-in-psycopg-3

PR #1340, opened 2021-08-19
Replace deprecated unittest aliases

hugovk:
https://docs.python.org/3/library/unittest.html#deprecated-aliases


dvarrazzo:
I like the deprecated aliases. They should be un-deprecated and the deprecators be deprecated.

hugovk:
Closing this year old PR, but happy to re-open if you change your mind 👍 

Note the deprecated aliases have been removed from Python 3.12, set for release in October 2023.

https://docs.python.org/3.12/whatsnew/3.12.html#removed

PR #1342, opened 2021-09-02
Docs > Build pre-reqs > Python header tweak

jecarr:
A small tweak in case anyone else comes across [this issue](https://stackoverflow.com/a/25723191).

(I kept getting the _Python.h: No such file or directory_ error when I thought I installed the correct package; turned out I needed one specific to my Python version)

Thanks!

### DIFF
--- a/doc/src/install.rst
+++ b/doc/src/install.rst
@@ -153,7 +153,8 @@ it from sources you will need:
 - A C compiler.
 
 - The Python header files. They are usually installed in a package such as
-  **python-dev** or **python3-dev**. A message such as *error: Python.h: No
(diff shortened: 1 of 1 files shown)


dvarrazzo:
That line is not a precise instruction for APT working in all the possible cases. it is only an indication that you need a way to find a `Python.h`.

 On Debian installing python3-dev is enough to install the dev package of the system Python: the dependencies take care of it. The answer on SO saying you require python-3.x-dev is wrong: you only need it if you install python 3.x manually.

jecarr:
Ok - thanks for the info @dvarrazzo!

PR #1353, opened 2021-09-21
Update pool methods with docs and elegant syntax

complexsum:


dvarrazzo:
What is the need of the elaborate docstring if it doesn't affect the docs at https://www.psycopg.org/docs/pool.html ?

complexsum:
@dvarrazzo I didn't knew about that but still it's good to have the docs updated in the code itself :)

dvarrazzo:
It is unneeded repetition IMO, unless you format that docstring in Sphinx and have the docs autogenerated from it.

Are you interested in this change? If not I will reject this as it is.

dvarrazzo:
Thank you for the contribution but it is past due date. Psycopg 3 pool is what is maintained now.

PR #1357, opened 2021-09-22
Add pg_config libdir to the library's rpath

### LINKED ISSUE #1356: Install script does not pass pg_config's LDFLAGS to the linker
The install script does not pass pg_config's LDFLAGS to the linker. In my case pg_config outputs the following LDFLAGS:

    LDFLAGS = -L/usr/lib64/llvm5.0/lib -L/usr/lib64 -Wl,--as-needed -Wl,-rpath,'/usr/pgsql-12/lib',--enable-new-dtags

But the psycopg2 install script ignores everything that doesn't start with "-L", therefore it ignores the rpath configuration. Then, when psycopg2 is loaded by a uwsgi executable that has its own rpath the process exits with a "missing name" error because the fact that the executable has its own rpath prevents it from loading the psycopg2 library even though ldconfig has run and has properly configured the system.

To work around this I do

    LDFLAGS='-L/usr/lib64/llvm5.0/lib -L/usr/lib64 -Wl,--as-needed -Wl,-rpath,/usr/pgsql-12/lib,--enable-new-dtags' pip install psycopg2

and then psycopg2 seems to work fine.

Please consider loading pg_config's LDFLAGS and passing them to the linker without user intervention.

gtzampanakis:
Close #1356.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,7 @@ Operating System :: Unix
 version_flags = ['dt', 'dec']
 
 PLATFORM_IS_WINDOWS = sys.platform.lower().startswith('win')
+PLATFORM_IS_LINUX = sys.platform.lower().startswith('linux')
 
 
 class PostgresConfig:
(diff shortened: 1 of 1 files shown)


gtzampanakis:
**NOTE:** I have tested this only on centos.

gtzampanakis:
Will this be merged?

dvarrazzo:
Hi, sorry, I thought you still had issues to solve around it.

This changeset makes me pretty nervous because it has just no test. Is there a way to reproduce the issue in first place and then to verify it's fixed by this change?

Can you comment in the setup.py why are we doing that thing?

Can you run a packaging task to verify that build works on all the platforms supported?

Thank you

zyv:
@gtzampanakis you should need to add the library path to `rpath` only if it's a dynamic library **and** the path to it is not already on the default linker path.

There are valid uses for `rpath`, but in this case you are 99% working around a broken installation of `libpq`, and unconditionally polluting `rpath` field for all users of the library.

One could do that conditionally by e.g. checking the output of `ldd`, but even this would be kind of questionable, because the problem shouldn't be there in the first place...

Issue #1174, opened 2020-10-21
ImportError: DLL load failed - Windows - pip install - python 3.9.0


adjenks:
I'm installing psycopg2 on Windows and I am getting this error.
I have pip 20.2.4
My python is version 3.9.0
I have the latest version of Visual studio installed along with the build tools.
I am not using pycharm. Just `pip install psycopg2`.
I have postgresql installed and all the PATH directories configured.
I built it and it failed to import with the error in a previous issue: https://github.com/psycopg/psycopg2/issues/1006
>ImportError: DLL load failed while importing _psycopg: The specified module could not be found.

dvarrazzo:
#1160

adjenks:
Thank you. I looked and found two other similar issues but not that one for some reason.

Issue #1160, opened 2020-10-09
ImportError: DLL load failed while importing _psycopg: The specified module could not be found in python 3.9


newETA:
Import failed while working with Python 3.9  experience this issue,  But it's fine in 3.7 and 3.8.

Using psycopg2 2.8.6

![image](https://user-images.githubusercontent.com/42071761/95525406-00573800-09a2-11eb-9156-8fbb54531500.png)

AntoninJcb:
I am having the same issue with python 3.9.
I could only make it work by reverting to 3.8.

dvarrazzo:
Python 3.9 is not supported yet. Packages have not been released.

hingston:
What is blocking the release?

dvarrazzo:
Appveyor doesn't support Python 3.9 yet: https://ci.appveyor.com/project/psycopg/psycopg2-wheels/builds/35677049

ahtik:
For reference, the Appveyor issue [tracking Python 3.9 support](https://github.com/appveyor/ci/issues/3541), which seems to prevent the (Python 3.9 Windows psycopg2-binary build)[https://pypi.org/project/psycopg2-binary/#files].

There seems to be a way to manually install Python 3.9 described in https://github.com/appveyor/ci/issues/3467#issuecomment-643534726. @dvarrazzo Could this be an option to get it built sooner?

dvarrazzo:
@ahtik I don't have resources to work on this issue in this moment. If someone provides a patch I can release packages before Appveyor completes Python 3.9 support.

ahtik:
@dvarrazzo Python 3.9.0 is now listed at https://github.com/appveyor/ci/issues/3541 and 3.9.0 being listed https://www.appveyor.com/docs/windows-images-software/#python as of today.

Could you try running the https://ci.appveyor.com/project/psycopg/psycopg2-wheels build again to heal the  PY_VER=39 section?

dvarrazzo:
@ahtik Sure, trying a build.

dvarrazzo:
@ahtik Still can't find the executable on the appveyor machine:

- https://ci.appveyor.com/project/psycopg/psycopg2/builds/35977860/job/u2ehf04x6l0wth03
- https://ci.appveyor.com/project/psycopg/psycopg2/builds/35977860/job/376li5550cfsjpcu

dvarrazzo:
[...]

Issue #1062, opened 2020-03-17
ImportError: DLL load failed while importing _psycopg: The specified module could not be found.


HTG-YT:
**Operating System:** Windows 10 Home
**IDE:** PyCharm Professional
**Python version:** 3.8.2
**pip version:** 19.0.3

I installed `psycopg2` with the command `pip install psycopg2==2.8.4`.

When I did `import psycopg2` inside my Python script, the error(in the title) occurred and following the error I have the problem right here:
```python
from psycopg2._psycopg import (                     # noqa
    BINARY, NUMBER, STRING, DATETIME, ROWID,

    Binary, Date, Time, Timestamp,
    DateFromTicks, TimeFromTicks, TimestampFromTicks,

    Error, Warning, DataError, DatabaseError, ProgrammingError, IntegrityError,
    InterfaceError, InternalError, NotSupportedError, OperationalError,

    _connect, apilevel, threadsafety, paramstyle,
    __version__, __libpq_version__,
)
```
So what happened here?

dvarrazzo:
Upgrade pip.

dvarrazzo:
Also, please open a bug to pycharm. I don't think I'm able to.

https://www.jetbrains.com/help/pycharm/reporting-issues.html#report-an-issue

HTG-YT:
Have upgraded pip to the latest version, the problem is still there.

HTG-YT:
@dvarrazzo and also opened an issue to pycharm. and one more thing the error seems to occur to the _psycopg2 PYD file

dvarrazzo:
See #1006. Please have this problem fixed in pycharm, because psycopg2 works alright with normal Python.

chrismahn2020:
> 
> 
> Have upgraded pip to the latest version, the problem is still there.

python -m pip install pip==19.3.1  should do the trick

dvarrazzo:
@chrismahn2020 19.3.1 is not the last pip version. Why are you suggesting to downgrade pip?

iamhpdev:
the only way to solve this error is downgrade the python version to second last recent version. for ex 10 is latest version then you have to install 9. and problem solved

adjenks:
I'm installing psycopg2 on Windows and I am getting this error.
I have pip 20.2.4
My python is version 3.9.0
I have the latest version of Visual studio installed along with the build tools.
I built it and it failed with the error in the previously linked issue: https://github.com/psycopg/psycopg2/issues/1006
[...]

PR #1362, opened 2021-10-06
Add note for dll import errors on Windows

bmarroquin:
I ran into this today and also noticed a few questions in closed issues related to this. I thought documenting this might help.

See:
#1174 
#1160 (some comments)
#1062


dvarrazzo:
Hello,

thank you for pointing this out: I wasn't aware of that function. However how does it apply to psycopg2? psycopg2 on windows packages its own libpq and doesn't use a system one. Does it mean that this applies not to final users, but to people who build psycopg2 on their own and uses either system libpq or a dll somewhere else?

PR #1374, opened 2021-10-17
make tox use expected python version

jouve:
Hi,

when running tox with the default envlist, it does not use the python I expect from tox.ini.

replacing `commands` with `python -V` in tox.ini, notice it's always Python 3.10.0:

```console
$ tox
GLOB sdist-make: /home/cyril/work/psycopg2/setup.py
3.6 create: /home/cyril/work/psycopg2/.tox/3.6
3.6 inst: /home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
3.6 installed: psycopg2 @ file:///home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
3.6 run-test-pre: PYTHONHASHSEED='231627133'
3.6 run-test: commands[0] | python -V
Python 3.10.0
3.7 create: /home/cyril/work/psycopg2/.tox/3.7
3.7 inst: /home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
3.7 installed: psycopg2 @ file:///home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
3.7 run-test-pre: PYTHONHASHSEED='231627133'
3.7 run-test: commands[0] | python -V
Python 3.10.0
...
```

but using the syntax described in [tox docs](https://tox.wiki/en/latest/config.html#tox-environments),
we get the expected behaviour:

```console
$ tox
GLOB sdist-make: /home/cyril/work/psycopg2/setup.py
py36 create: /home/cyril/work/psycopg2/.tox/py36
ERROR: InterpreterNotFound: python3.6
py37 create: /home/cyril/work/psycopg2/.tox/py37
ERROR: InterpreterNotFound: python3.7
py38 create: /home/cyril/work/psycopg2/.tox/py38
ERROR: InterpreterNotFound: python3.8
py39 create: /home/cyril/work/psycopg2/.tox/py39
py39 inst: /home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
py39 installed: psycopg2 @ file:///home/cyril/work/psycopg2/.tox/.tmp/package/1/psycopg2-2.9.1.zip
py39 run-test-pre: PYTHONHASHSEED='487652293'
py39 run-test: commands[0] | python -V
Python 3.9.7
py310 create: /home/cyril/work/psycopg2/.tox/py310
[...]

PR #1375, opened 2021-10-18
get rid of tox

jouve:
https://github.com/psycopg/psycopg2/pull/1374


jouve:
to clarify:

the point of tox is that, for each entry in `[tox]envlist`, it creates a virtualenv and run the `[testenv] commands` inside.

for ex, running `tox` would run the commands in `[testenv]` once for each entry in envlist.
To specify different python version in the envlist, then needs to be named pyNM (N = 3, M = 9).

In this repo, the envlist is not generated properly (env named 3.x instead of py3x), so it just reuse the default python on the system.

If any user is using tox to test, then it's not doing what they expect.

jouve:
yes you used the good syntax :)

the one is tox.ini is the wrong one

PR #1376, opened 2021-10-18
python 3.11 support

jouve:
preliminary [python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html) support:

* The <Python.h> header file no longer includes <stdlib.h>. C extensions using <stdlib.h> must now include it explicitly. The system <stdlib.h> header provides functions like: malloc()/free(), getenv(), strtol(), abs(), strtol(), exit() and abort(). (Contributed by Victor Stinner in bpo-45434.)
* The PyType_Ready() function now raises an error if a type is defined with the Py_TPFLAGS_HAVE_GC flag set but has no traverse function (PyTypeObject.tp_traverse). (Contributed by Victor Stinner in bpo-44263.)

### DIFF
--- a/.appveyor/packages.yml
+++ b/.appveyor/packages.yml
@@ -11,6 +11,8 @@ environment:
     matrix:
         # For Python versions available on Appveyor, see
         # https://www.appveyor.com/docs/windows-images-software/#python
+        - {APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019, PY_VER: "310", PY_ARCH: "32"}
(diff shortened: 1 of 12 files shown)


dvarrazzo:
Hello, thank you for this changeset.

Could you please leave the tox changes out of this changeset and only change the code and the pipeline run?

jouve:
If you think the changes to tests.yml are too much just to support an alpha release of python 3.11, I'll agree :)

jouve:
@dvarrazzo it got closed because you delete py310 branch, should I reopen on master or keep closed ?

dvarrazzo:
Rebase it on master :)

PR #1392, opened 2021-11-18
Build `musllinux` wheels

lithammer:
I'm back for some more `musllinux` support. Since psycopg2 is still needed for Django, I figured I'd try to add support here as well.

The artifacts can be found here: https://github.com/lithammer/psycopg2/actions/runs/1564148387

### DIFF
--- a/scripts/build/build_musllinux_1_1.sh
+++ b/scripts/build/build_musllinux_1_1.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+# Create musllinux_1_1 wheels for psycopg2
+#
+# Look at the .github/workflows/packages.yml file for hints about how to use it.
+
+set -euo pipefail
+set -x
+
+dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+prjdir="$( cd "${dir}/../.." && pwd )"
+
+# Build all the available versions, or just the ones specified in PYVERS
+if [ ! "${PYVERS:-}" ]; then
+    PYVERS="$(ls /opt/python/)"
+fi
+
+# Find psycopg version
+version=$(grep -e ^PSYCOPG_VERSION "${prjdir}/setup.py" | sed "s/.*'\(.*\)'/\1/")
+# A gratuitous comment to fix broken vim syntax file: '")
+distdir="${prjdir}/dist/psycopg2-$version"
+
+# Replace the package name
+if [[ "${PACKAGE_NAME:-}" ]]; then
+    sed -i "s/^setup(name=\"psycopg2\"/setup(name=\"${PACKAGE_NAME}\"/" \
+        "${prjdir}/setup.py"
+fi
+
+# Install prerequisite libraries
+apk update
+apk add postgresql-dev
+# Add findutils because the Busybox version lacks the `-ls` flag, used by the
+# `strip_wheel.sh` script.
+apk add findutils
+
+# Create the wheel packages
+for pyver in $PYVERS; do
+    pybin="/opt/python/${pyver}/bin"
+    "${pybin}/python" -m build -w -o "${prjdir}/dist/" "${prjdir}"
+done
+
+# Bundle external shared libraries into the wheels
+for whl in "${prjdir}"/dist/*.whl; do
+    "${dir}/strip_wheel.sh" "$whl"
+    auditwheel repair "$whl" -w "$distdir"
+done
+
+# Make sure the postgresql-dev is not in the system
+apk del postgresql-dev
+
+# Install packages and test
+cd "${prjdir}"
+for pyver in $PYVERS; do
+    pybin="/opt/python/${pyver}/bin"
+    "${pybin}/pip" install ${PACKAGE_NAME:-psycopg2} --no-index -f "$distdir"
+
+    # Print psycopg and libpq versions
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.__version__)"
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.__libpq_version__)"
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.extensions.libpq_version())"
+
+    # Fail if we are not using the expected libpq library
+    if [[ "${WANT_LIBPQ:-}" ]]; then
+        "${pybin}/python" -c "import psycopg2, sys; sys.exit(${WANT_LIBPQ} != psycopg2.extensions.libpq_version())"
+    fi
+
+    "${pybin}/python" -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
+done
--- a/.github/ISSUE_TEMPLATE/problem-installing-psycopg2.md
+++ b/.github/ISSUE_TEMPLATE/problem-installing-psycopg2.md
@@ -14,7 +14,6 @@ If you have a question, such has "how do you do X with Python/PostgreSQL/psycopg
 - [ ] I am running the latest version of pip, i.e. typing ``pip --version`` you get [this version](https://pypi.org/project/pip/).
 - [ ] I have read the [installation documentation](https://www.psycopg.org/docs/install.html) and the [frequently asked questions](https://www.psycopg.org/docs/faq.html)
 - [ ] I am not trying to install `psycopg2-binary` on an Apple M1, for which [binary packages are not supported](https://github.com/psycopg/psycopg2/issues/1286)
-- [ ] I am not trying to install `psycopg2-binary` in an Alpine Linux Docker image
 - [ ] If install failed, I typed `pg_config` on the command line and I obtained an output instead of an error.
 
 **Please complete the following information:**
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -61,6 +61,10 @@ jobs:
           - {tag: manylinux2014, arch: i686}
           - {tag: manylinux_2_24, arch: aarch64}
           - {tag: manylinux_2_24, arch: ppc64le}
+          - {tag: musllinux_1_1, arch: x86_64}
+          - {tag: musllinux_1_1, arch: i686}
+          - {tag: musllinux_1_1, arch: aarch64}
+          - {tag: musllinux_1_1, arch: ppc64le}
 
     runs-on: ubuntu-20.04
     steps:
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 Current release
 ---------------
 
+What's new in psycopg 2.9.3
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Alpine (musl) wheels now available (:ticket:`#1148`).
+
 What's new in psycopg 2.9.2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 


scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
lithammer, line 1: This script is more or less a copy of `scripts/build/build_manylinux_2_24.sh` some minor tweaks:

```diff
diff --git a/scripts/build/build_manylinux_2_24.sh b/scripts/build/build_musllinux_1_1.sh
index 6a9e992b..90ad9cca 100755
--- a/scripts/build/build_manylinux_2_24.sh
+++ b/scripts/build/build_musllinux_1_1.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Create manylinux_2_24 wheels for psycopg2
+# Create musllinux_1_1 wheels for psycopg2
 #
 # Look at the .github/workflows/packages.yml file for hints about how to use it.
 
@@ -27,16 +27,13 @@ if [[ "${PACKAGE_NAME:-}" ]]; then
 fi
 
 # Install prerequisite libraries
-curl -k -s https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
-echo "deb http://apt.postgresql.org/pub/repos/apt stretch-pgdg main" \
-    > /etc/apt/sources.list.d/pgdg.list
-apt-get -y update
-apt-get install -y libpq-dev
+apk update
+apk add postgresql-dev tzdata
 
 # Create the wheel packages
 for pyver in $PYVERS; do
     pybin="/opt/python/${pyver}/bin"
-    "${pybin}/pip" wheel "${prjdir}" -w "${prjdir}/dist/"
+    "${pybin}/python" -m build -w -o "${prjdir}/dist/" "${prjdir}"
 done
 
 # Bundle external shared libraries into the wheels
@@ -44,12 +41,6 @@ for whl in "${prjdir}"/dist/*.whl; do
     auditwheel repair "$whl" -w "$distdir"
 done
 
-# Make sure the libpq is not in the system
-for f in $(find /usr/lib /usr/lib64 -name libpq\*) ; do
-    mkdir -pv "/libpqbak/$(dirname $f)"
-    mv -v "$f" "/libpqbak/$(dirname $f)"
-done
-
 # Install packages and test
 cd "${prjdir}"
 for pyver in $PYVERS; do
@@ -68,8 +59,3 @@ for pyver in $PYVERS; do
 
     "${pybin}/python" -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
 done
-
-# Restore the libpq packages
-for f in $(cd /libpqbak/ && find . -not -type d); do
-    mv -v "/libpqbak/$f" "/$f"
-done
```

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+# Install prerequisite libraries
+apk update
+apk add postgresql-dev tzdata
+
+# Create the wheel packages
+for pyver in $PYVERS; do
+    pybin="/opt/python/${pyver}/bin"
+    "${pybin}/python" -m build -w -o "${prjdir}/dist/" "${prjdir}"
```
lithammer, line 36: Using [pypa/build](https://github.com/pypa/build) here instead of `pip wheel`. Which is basically the future.

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,69 @@
...
+    PYVERS="$(ls /opt/python/)"
+fi
+
+# Don't build musl wheels for Python 3.9 and 3.10 which is the version on
+# Alpine 3.14 and 3.15. The will install but cause 'ImportError' because of
+# conflicting SOAPI values (wanting '-musl.so' but getting '-gnu.so').
+#
+# https://discuss.python.org/t/a-mess-with-soabi-tags-on-musllinux/11688
+PYVERS=${PYVERS//cp39-cp39/}
+PYVERS=${PYVERS//cp310-cp310/}
```
lithammer, line 24: This is a bit unfortunate and I suppose something similar should be done for psycopg3. But in my opinion it's better to fail early during installation than at runtime.

Alpine 3.15 seems to actually be using Python 3.9 still, but I think it's probably a good idea to be a bit proactive in case they decide to update to 3.10.

lithammer: Other relevant links:

- https://github.com/nedbat/coveragepy/issues/1268
- https://github.com/python/cpython/pull/24502

dvarrazzo:
Uhm, on psycopg 3 we build for 3.9 and 3.10 too, then we run tests, which pass. Do you think we should take a closer look at them? We should check the libraries bundled with the binary packages and we can compare the times to pass the test version-to-version to check if we are affected by a regression too. But that's a psycopg 3 matter.

Thinking about psycopg2 instead: do we want to switch to cibuildwheel instead?

scripts/build/build_musllinux_1_1.sh
lithammer: Seems like this is getting fixed upstream in Alpine and backported to at least 3.15 (hopefully 3.14 as well). How do you want to approach this from a psycopg2 perspective? Just revert this workaround and maybe wait a while before merging to let the Alpine patch to propagate. Or keep the workaround and revert it sometime in the future?

https://gitlab.alpinelinux.org/alpine/aports/-/issues/13227

lithammer:
> Uhm, on psycopg 3 we build for 3.9 and 3.10 too, then we run tests, which pass. Do you think we should take a closer look at them? We should check the libraries bundled with the binary packages and we can compare the times to pass the test version-to-version to check if we are affected by a regression too. But that's a psycopg 3 matter.

Since it seems like this will be fixed in Alpine in due time, maybe no action is needed for now...?

> Thinking about psycopg2 instead: do we want to switch to cibuildwheel instead?

That's probably a good idea. That way it's easier to take advantage of whatever fixes they make to the build process (i.e. just upgrade the GitHub Action).

Do you still want this change meanwhile or should I close?

scripts/build/build_musllinux_1_1.sh
lithammer: Hmm, seems like it has already been backported to 3.14 and 3.15 🥳

dvarrazzo: > Hmm, seems like it has already been backported to 3.14 and 3.15 :partying_face: 


Yes, tested with `psycopg[binary]` and fresh alpine images and works as expected. I'd say this is no more an issue.

lithammer:
> > Do you still want this change meanwhile or should I close it?
> 
> It's up to you. Copying and adapting the existing script seems a much easier work. I personally don't wish to spend a lot of work converting the pipeline to cibuildwheel, but I don't want to stop you if you think that it would make easier to generate musl wheels.

Ultimately I would like to try and convert to cibuildwheel. But it's hard to say when I'll find the motivation, so I think it's better to merge this now instead of waiting indefinitely for musllinux support.

lithammer:
I've removed the Alpine 3.14-3.15 hack now as well.

lithammer:
Gentle nudge @dvarrazzo ☝️

scripts/build/build_musllinux_1_1.sh
dvarrazzo:
```diff
-# Make sure the libpq is not in the system
```
Can we restore these lines? They check that the binary packages work even without the system libpq and I'd like to keep that check.

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Create the wheel packages
+for pyver in $PYVERS; do
+    pybin="/opt/python/${pyver}/bin"
+    "${pybin}/python" -m build -w -o "${prjdir}/dist/" "${prjdir}"
+done
+
+# Bundle external shared libraries into the wheels
+for whl in "${prjdir}"/dist/*.whl; do
+    auditwheel repair "$whl" -w "$distdir"
```
dvarrazzo, line 41: Is this a good moment to compress the library, using the script already available?

scripts/build/build_musllinux_1_1.sh
dvarrazzo: I am not aware, but if it passes the tests, good for me :)

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,61 @@
...
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.__version__)"
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.__libpq_version__)"
+    "${pybin}/python" -c "import psycopg2; print(psycopg2.extensions.libpq_version())"
+
+    # Fail if we are not using the expected libpq library
+    if [[ "${WANT_LIBPQ:-}" ]]; then
+        "${pybin}/python" -c "import psycopg2, sys; sys.exit(${WANT_LIBPQ} != psycopg2.extensions.libpq_version())"
+    fi
+
+    "${pybin}/python" -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
```
dvarrazzo, line 60: I understand that you have deleted the instruction to hide away the system libpq:

```diff
@@ -44,12 +49,6 @@
     auditwheel repair "$whl" -w "$distdir"
 done
 
-# Make sure the libpq is not in the system
-for f in $(find /usr/lib /usr/lib64 -name libpq\*) ; do
-    mkdir -pv "/libpqbak/$(dirname $f)"
-    mv -v "$f" "/libpqbak/$(dirname $f)"
-done
-
 # Install packages and test
 cd "${prjdir}"
 for pyver in $PYVERS; do
@@ -68,8 +67,3 @@
 
     "${pybin}/python" -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
 done
-
-# Restore the libpq packages
-for f in $(cd /libpqbak/ && find . -not -type d); do
-    mv -v "/libpqbak/$f" "/$f"
-done
```

can we restore them please? The aim is to test that the binary packages work in a system without the libpq installed.

Alternatively you can do an `apk delete postgresql-dev`. We don't really care about having the package after build.

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Replace the package name
+if [[ "${PACKAGE_NAME:-}" ]]; then
+    sed -i "s/^setup(name=\"psycopg2\"/setup(name=\"${PACKAGE_NAME}\"/" \
+        "${prjdir}/setup.py"
+fi
+
+# Install prerequisite libraries
+apk update
+apk add postgresql-dev tzdata
```
dvarrazzo, line 31: I think `tzdata` is something we needed for psycopg 3. Or do tests break on psycopg2 too?

dvarrazzo (changes requested):
I think we should amend the docs somewhere (install? readme?) to declare Alpine linux supported. Please grep around for musl or alpine to check if we can adjust something.

Can you also add an entry to the news file to record that Alpine packages are to be added from 2.9.3?

dvarrazzo:
> Gentle nudge @dvarrazzo point_up

Hi, sorry, I was of the impression that there was still something to do here. Sent you a review with something to fix, then happy to release those binaries.

lithammer:
> > Gentle nudge @dvarrazzo ☝️
> 
> Hi, sorry, I was of the impression that there was still something to do here. Sent you a review with something to fix, then happy to release those binaries.

No worries! I'll look into fixing the review comments.

scripts/build/strip_wheel.sh
```diff
@@ -32,6 +32,10 @@ trap "rm -r ${tmpdir}" EXIT
 cd "${tmpdir}"
 $py -m zipfile -e "${wheel}" .
 
+# Add findutils because the Busybox version lacks the `-ls` flag.
+. /etc/os-release
+[[ "$ID" = 'alpine' ]] && apk add findutils
```
lithammer, line 37:
```console
$ find . -name '*.so' -ls -exec strip '{}' ';'
find: unrecognized: -ls
```

dvarrazzo, line 37: Could you please have this in build_musllinux please? It's bad to have it here, especially so down in the script

lithammer:
I've fixed everything except this 👇 I couldn't really find a decent place for it. I tried seeing if there was explicit mention of Windows support but couldn't really find something like that either (except for maybe the build status at the bottom of the README).

> I think we should amend the docs somewhere (install? readme?) to declare Alpine linux supported. Please grep around for musl or alpine to check if we can adjust something.

The wheels can found [here](https://github.com/lithammer/psycopg2/actions/runs/1564148387).

PS: Stripping the `.so` file reduced the size from `1490488` to `313760` on `x86_64`. Which is quite substantial.

scripts/build/build_musllinux_1_1.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
lithammer, line 1:
```diff
diff --git a/scripts/build/build_manylinux_2_24.sh b/scripts/build/build_musllinux_1_1.sh
index 742014ad..62860dcb 100755
--- a/scripts/build/build_manylinux_2_24.sh
+++ b/scripts/build/build_musllinux_1_1.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Create manylinux_2_24 wheels for psycopg2
+# Create musllinux_1_1 wheels for psycopg2
 #
 # Look at the .github/workflows/packages.yml file for hints about how to use it.
 
@@ -27,16 +27,16 @@ if [[ "${PACKAGE_NAME:-}" ]]; then
 fi
 
 # Install prerequisite libraries
-curl -k -s https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
-echo "deb http://apt.postgresql.org/pub/repos/apt stretch-pgdg main" \
-    > /etc/apt/sources.list.d/pgdg.list
-apt-get -y update
-apt-get install -y libpq-dev
+apk update
+apk add postgresql-dev
+# Add findutils because the Busybox version lacks the `-ls` flag, used by the
+# `strip_wheel.sh` script.
+apk add findutils
 
 # Create the wheel packages
 for pyver in $PYVERS; do
     pybin="/opt/python/${pyver}/bin"
-    "${pybin}/pip" wheel "${prjdir}" -w "${prjdir}/dist/"
+    "${pybin}/python" -m build -w -o "${prjdir}/dist/" "${prjdir}"
 done
 
 # Bundle external shared libraries into the wheels
@@ -45,11 +45,8 @@ for whl in "${prjdir}"/dist/*.whl; do
     auditwheel repair "$whl" -w "$distdir"
 done
 
-# Make sure the libpq is not in the system
-for f in $(find /usr/lib /usr/lib64 -name libpq\*) ; do
-    mkdir -pv "/libpqbak/$(dirname $f)"
-    mv -v "$f" "/libpqbak/$(dirname $f)"
-done
+# Make sure the postgresql-dev is not in the system
+apk del postgresql-dev
 
 # Install packages and test
 cd "${prjdir}"
@@ -69,8 +66,3 @@ for pyver in $PYVERS; do
 
     "${pybin}/python" -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
 done
-
-# Restore the libpq packages
-for f in $(cd /libpqbak/ && find . -not -type d); do
-    mv -v "/libpqbak/$f" "/$f"
-done
```

lithammer:
I have no more changes planned.

dvarrazzo:
I apologise, I lost track of this task. Building test packages and merging.

lithammer:
No worries! These are busy times of the year. Thanks for taking your time.

PR #1393, opened 2021-11-21
Fix typo in large object usage documentation

wjt:


dvarrazzo:
Contidions, of course! :smile: Thank you

PR #1484, opened 2022-07-30
Remove `__nonzero__` method

sobolevn:
It is not used in python3, but python2 is no longer supported: https://github.com/python-babel/babel/blob/d8584405a66a79d79bed44b5d80df0bc89f8eee3/setup.py#L59


dvarrazzo:
Thank you, merged 🙂

PR #834, opened 2019-01-16
added helpful executemany & execute_batch examples

kannes:
hopefully this will make those functions more approachable for newbies. everyone tries it without proper sequences and then has to google.

PR #1508, opened 2022-10-10
Add executemany & execute_batch examples

kannes:
I had somehow forgotten about https://github.com/psycopg/psycopg2/pull/834 and could not fix the issues in that branch as I have no access to it somehow... Here is a new PR with the requested changes. I don't remember if I tested building the docs with this back in 2019, the syntax for the `extras.rst` might be broken but I know nothing about RST.

My original text: "hopefully this will make those functions more approachable for newbies. everyone tries it without proper sequences and then has to google."

### DIFF
--- a/doc/src/cursor.rst
+++ b/doc/src/cursor.rst
@@ -207,10 +207,12 @@ The ``cursor`` class
 
         Parameters are bounded to the query using the same rules described in
         the `~cursor.execute()` method.
-        
+
+        .. code:: python
+
             >>> nums = ((1,), (5,), (10,))
             >>> cur.executemany("INSERT INTO test (num) VALUES (%s)", nums)
-            
+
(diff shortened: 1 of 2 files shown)


dvarrazzo:
This doesn't seem valid reST. Have you generated the docs? I bet you get warnings or error.

I think you need a `..code:: python` in front of those examples, or a `::` on a line of its own.

kannes:
Thanks for taking a look! I imitated other code examples in those files and they seem to build fine:

image
image

dvarrazzo:
Let's say that Sphinx is doing what you mean instead than what you say. It seems it converted what in reST is a block quote into a doctest.

I have fixed the syntax and the whitespaces, because I'm an old grumpy poo, and committed your changes. Thank you very much for your contribution! :slightly_smiling_face:

kannes:
Awesome! Thank you :)

Issue #1506, opened 2022-10-05
Regression test failure with PG15: test_ssl_attribute AssertionError: 'OpenSSL' is not None


df7cb:
Hi, running the regression tests for psycopg2 2.9.3 on Debian unstable against PG15rc2 yields this:

```
LANG=C.UTF-8 LC_ALL=C.UTF-8 PSYCOPG2_TESTDB_HOST=localhost PSYCOPG2_TESTDB=postgres PSYCOPG2_TEST_FAST=1 pg_virtualenv -v 15 python3.10 -c import tests; tests.unittest.main(defaultTest='tests.test_suite') --verbose
...
16:36:40 test_server_version (tests.test_connection.TestConnectionInfo) ... ok
16:36:40 test_socket (tests.test_connection.TestConnectionInfo) ... ok
16:36:40 test_ssl_attribute (tests.test_connection.TestConnectionInfo) ... FAIL
16:36:40 test_ssl_in_use (tests.test_connection.TestConnectionInfo) ... ok
16:36:40 test_ssl_not_supported (tests.test_connection.TestConnectionInfo) ... skipped 'skipped because libpq 140004'
...
16:36:56 test_subclass (tests.test_with.WithCursorTestCase) ... ok
16:36:56 test_with_error (tests.test_with.WithCursorTestCase) ... ok
16:36:56 test_with_ok (tests.test_with.WithCursorTestCase) ... ok
16:36:56 
16:36:56 ======================================================================
16:36:56 FAIL: test_ssl_attribute (tests.test_connection.TestConnectionInfo)
16:36:56 ----------------------------------------------------------------------
16:36:56 Traceback (most recent call last):
16:36:56   File "/tmp/autopkgtest.PyQGhl/build.UJR/src/tests/test_connection.py", line 1936, in test_ssl_attribute
16:36:56     self.assertIsNone(self.bconn.info.ssl_attribute(attrib))
16:36:56 AssertionError: 'OpenSSL' is not None
16:36:56 
16:36:56 ----------------------------------------------------------------------
16:36:56 Ran 738 tests in 19.971s
16:36:56 
16:36:56 FAILED (failures=1, skipped=94)
```

Full log: https://pgdgbuild.dus.dg-i.net/job/psycopg2-autopkgtest-beta/architecture=amd64,distribution=sid/1/console

A new release with a fix would be appreciated so we can update the package without triggering regressions when PG15 is uploaded.

dvarrazzo:
[...]

Issue #1521, opened 2022-11-11
Regression test failure with PG 15.1: tests/test_connection.py: AssertionError: [] is not true


df7cb:
It looks that #1506 is back - Pycopg2 2.9.5, PostgreSQL 15.1, Debian unstable/amd64:

```
FAIL: test_ssl_attribute (tests.test_connection.TestConnectionInfo)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/cbe/projects/postgresql/psycopg2/psycopg2/.pybuild/cpython3_3.10_psycopg2/build/tests/test_connection.py", line 1925, in test_ssl_attribute
    self.assert_(attribs)
AssertionError: [] is not true
```

The culprit seems to be postgres/postgres@cae4688ce8.

dvarrazzo:
This area is being worked up quite a lot. Could you please suggest how to fix these testt in a way that won't be broken by upstream in two weeks? Should we drop them?

df7cb:
TBH I don't really know. Given it's a 15.1 change, it's probably less likely that it will happen again in a higher minor release. I'd probably update to the current format, plus whatever the old releases produced, and forget that 15.0 existed.

dvarrazzo:
Actually I don't see the relation between that changeset and this error.

The test failing

https://github.com/psycopg/psycopg2/blob/e8d92b74fd4b219cb4725cb906c526f3f2dce31d/tests/test_connection.py#L1923-L1925

relates to the function `PQsslAttributeNames()`, which doesn't seem touched by the changeset you mention. I don't know how to reproduce getting an empty list from that function:
```python
>>> import psycopg2

>>> psycopg2._psycopg.libpq_version()
150001

>>> conn = psycopg2.connect("")
>>> conn.info.ssl_in_use
False

>>> conn.info.ssl_attribute_names
['library', 'key_bits', 'cipher', 'compression', 'protocol']
```

I get a list even with a broken connection.

How is the psycopg2 in your tests built? Maybe with a libpq with no ssl support?

df7cb:
Hmm, the tests are passing with 15.x now, so I'll just close this for now.

For PG16, see #1582.

dvarrazzo:
Assumed fixed by #1582. If this is a problem for 15.1 only it's not worth being fixed.

PR #1582, opened 2023-05-29
Disable test_ssl_attribute on PG16+

df7cb:
PG15 changed the semantics of some ssl attributes (#1506), and a very similar regression test failure has now been observed again with PG16. Disable the test for now.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1920,6 +1920,7 @@ class TestConnectionInfo(ConnectingTestCase):
             self.conn.info.ssl_attribute('wat')
 
     @skip_before_libpq(9, 5)
+    @skip_after_libpq(16)
     def test_ssl_attribute(self):
         attribs = self.conn.info.ssl_attribute_names
         self.assert_(attribs)


tests/test_connection.py
```diff
@@ -1920,6 +1920,7 @@ def test_ssl_not_supported(self):
             self.conn.info.ssl_attribute('wat')
 
     @skip_before_libpq(9, 5)
+    @skip_after_libpq(16)
```
dvarrazzo, line 1923: I thought you found some probems in 15.1 too? See #1521 

Maybe it would be safer

```suggestion
    @skip_after_libpq(15, 1)
```

or just `15`?

df7cb:
The current test works with 15.2++, so I left that in place. Iirc the outcome back then was that the workaround fixed the issue for anything but 15.0/15.1 and we'd just ignore the problem with these earlier versions since they are unlikely to be encountered in the wild. If we make it 15,1, we could revert 6d815f5df95b, but then we'd lose test coverage of 15.1++.

df7cb:
Or maybe I'm mixing that up with some other 15.0/.1 change, this seems to have been rc2... anyway, the test passes now for 15.current.

dvarrazzo:
Ok, let's assume it was a 15.1 problem only. Merging it this way, but feel free to keep on tweaking if you see the need. Thank you very much :slightly_smiling_face:

PR #1622, opened 2023-09-19
AbstractConnectionPool does not pool more than "minconn" connections.

mjgp2:
The pool should add the connection back into the pool as long as the pool is not larger than the maximum pool size.

The current code means that the reusable pool can never get bigger than "minconn". 

For example, if you want a pool that has zero minimum connections in it, there will never be a connection reused.


mjgp2:
It seems like this is the expected behaviour, which is very strange.

PR #1631, opened 2023-10-06
Show a Changelog link on PyPI

takis:


dvarrazzo:
Thank you, this is a good idea. However the proper public url for the changelog is `https://www.psycopg.org/docs/news.html` (which is a rendered version of the `NEWS` file). 

Would you like to fix your MR so I'll keep your attribution?

Thank you very much!

dvarrazzo:
Would it render the right icon using the "Changes" entry? Looking at Pydantic, for instance, [it uses the entry "Changelog"](https://github.com/pydantic/pydantic/blob/32ea570bf96e84234d2992e1ddf40ab8a565925a/pyproject.toml#L130C1-L130C10).

dvarrazzo:
Well, I'm silly, [psycopg 3 uses "Changes"](https://github.com/psycopg/psycopg/blob/2b488bf728977276ffd74dd40bb0d6132f9d10c6/psycopg/setup.cfg#L23) and it works ok, so that'll be fine....

takis:
Hi Daniele,

Yes, I used 'Changes'  for similarity to the psycopg3 `pyproject.toml` file.

I updated the URL to the one you specified.

With friendly regards,
Takis

PR #1639, opened 2023-10-29
chore: update GitHub actions and Postgres image tags

reneleonhardt:
After updating all GitHub actions I saw that old Postgres versions were used, so I updated them to 16 as well.

* build-sdist used 13
* build-linux used 14 (but libpq 16)


dvarrazzo:
Thank you for the cleanup. The postgres you changed are only used for quick functionality checks of the packages, not for the complete test suite, but it's good to keep them up to date.

PR #1640, opened 2023-10-29
chore: Let dependabot update GitHub actions

reneleonhardt:
Let bots do some of the tedious work 😄

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "weekly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
```
dvarrazzo, line 3: Which dependencies would this end up updating?

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
dvarrazzo, line 10: Please add a newline at the end of the file.

.github/dependabot.yml
reneleonhardt: As far as I can see only doc/requirements.txt:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile
But the PRs will not be auto-merged, you still have to merge the suggestions yourself, but you don't have to update manually anymore 😅

dvarrazzo: Yes, I'm not a heavy user of dependabot, but many projects I maintain still receive security-oriented upgrade suggestions.

However with the docs is tricky: we don't have automatic tests that verify that the documentation is rendered correctly. Merging the change would generate new docs, which might be broken. So I don't think I want to run the risk.

Can you please limit the updates to the github actions, and set it to a monthly update? I don't think a weekly update is requested.

reneleonhardt: Sure thing.

PR #1668, opened 2024-02-02
Document bytes as another valid input type for lobject.write()

nickzandbergen:
I'm not sure what the correct syntax should be, since 3.9 uses Union[t1, t2] for representing multiple types.


dvarrazzo:
That documentation predates the Python types annotation, so it most places it would be informal (e.g. `write(str or bytes)`. What you propose is well understandable, it would be ok for me to use.

nickzandbergen:
Thanks for letting me know, I'll mark this as ready for review as-is then.

dvarrazzo:
Thank you very much!

Issue #1670, opened 2024-02-05
LoggingConnection keeps connections open even after the Garbage Collector is run


spatankar-dmpna:
**Please complete the following information:**
- OS: Windows
- Psycopg version: 2.8.3
- Python version: 3.6
- PostgreSQL version: 13
- pip version: conda 23.7.4

**Describe the bug**
Please let us know:

1: what you did
We have subclassed LoggingConnection and LoggingCursor for our project. In our initial design we were relying on the garbage collectors to close our connection at the end of the function scopes. This was bad design and has since been fixed.
2: what you expected to happen
The garbage collector should have destroyed the LoggingConnection object at the end the function scope.
3: what happened instead
The LoggingConnection object remained alive because it still had a reference to it.

Explanation:

Looking into the issue, it looks like the LoggingConnection still had a reference to it because of the bound method "log"

`def initialize(self, logobj):
        """Initialize the connection to log to `!logobj`.

        The `!logobj` parameter can be an open file object or a Logger/LoggerAdapter
        instance from the standard logging module.
        """
        self._logobj = logobj
        if _logging and isinstance(
                logobj, (_logging.Logger, _logging.LoggerAdapter)):
            self.log = self._logtologger
        else:
            self.log = self._logtofile`

The Solution:

The methods should be bound with a weak reference instead

`
def initialize(self, logobj):
        """Initialize the connection to log to `!logobj`.

        The `!logobj` parameter can be an open file object or a Logger/LoggerAdapter
        instance from the standard logging module.
        """
        self._logobj = logobj
        if _logging and isinstance(
                logobj, (_logging.Logger, _logging.LoggerAdapter)):
            self._log = weakref.WeakMethod(self._logtologger)
        else:
            self._log = weakref.WeakMethod(self._logtofile)
            
def log(self, *args, **kwargs):
         return self._log()(*args, **kwargs)
`

PR #1671, opened 2024-02-06
Made the log method a weakreference

spatankar-dmpna:
This is a fix for [issue 1670](https://github.com/psycopg/psycopg2/issues/1670). The issue outlines a bug where garbage collector cannot destroy the LoggingConnection object since it has a bound method called log on it.

### DIFF
--- a/lib/extras.py
+++ b/lib/extras.py
@@ -409,7 +409,9 @@ class LoggingConnection(_connection):
         """
         Public interface of the log method defined in initialize
         """
-        return self._log()(*args, **kwargs)
+        log = self._log()
+        if log:
+            return self._log()(*args, **kwargs)
 
     def filter(self, msg, curs):
         """Filter the query before logging it.


lib/extras.py
```diff
@@ -401,9 +401,15 @@ def initialize(self, logobj):
...
+            self._log = _weakref.WeakMethod(self._logtologger)
         else:
-            self.log = self._logtofile
+            self._log = _weakref.WeakMethod(self._logtofile)
+
+    def log(self, *args, **kwargs):
+        """
+        Public interface of the log method defined in initialize
+        """
+        return self._log()(*args, **kwargs)
```
dvarrazzo, line 412: You should check if `self._log()` returns None. In that case I think you can just plonk the message: it likely means that the program is being terminated.

spatankar-dmpna: So you mean something like this?
`

    if self._log:
         return self._log()(*args, **kwargs)

`

dvarrazzo: No, more:

```python
log = self._log()
if log:
    log(*args, **kwargs)
```

spatankar-dmpna: I'm not a 100% sure of this myself but wouldn't this replicate the same bound method issue again?

It is possible that the WeaktMethod does decrease the reference count even after we do this though.

dvarrazzo: > I'm not a 100% sure of this myself but wouldn't this replicate the same bound method issue again?

No, because this would be into the `log()` function. `log`, here, is a local variable.

spatankar-dmpna: I also wanted to ask if this issue might be happening in other parts. I could make the same changes there too.

dvarrazzo:
I played a bit more with the issue. I didn't realise that the pattern of copying a bound method to the state creates a loop. It might be somewhere else too, sure; I have in mind psycopg 3 date/time adapters; I don't know about psycopg2.

However, looking at it, I don' think I would use weakref at all. ISTM that the same can be achieved by taking a class method reference in the state, instead of the bound method. Something like, untested:

```python
        if _logging and isinstance(
                logobj, (_logging.Logger, _logging.LoggerAdapter)):
            self._log_method = type(self)._logtologger
        else:
            self._log_method = type(self)._logtofile

    def log(self, msg, curs):
        """
        Public interface of the log method defined in initialize
        """
        self._log_method(self, msg, curs)
```

dvarrazzo:
...or, even cleaner, make _logtologger, _logtofile two staticmethods (taking a `self` as first parameter anyway) and avoid taking them from `type(self)` in the constructor.

dvarrazzo:
See https://github.com/psycopg/psycopg/commit/36b0a5506bb4fec70ca7cb9fa38a88c2e73a094d for an example of the same problem addressed in psycopg 3.

dvarrazzo:
As for whether there are similar cases in psycopg2, using `ag 'self\..* = self\.'` I could only see the pattern used in the hstore:

https://github.com/psycopg/psycopg2/blob/00870545b72231775258af4a4dd40842d2f18f0f/lib/extras.py#L797-L798

It's worth fixing it for consistency, however that case only happens with hstore before Postgres 9.0 it seems, so only with long unsupported Postgres versions.

spatankar-dmpna:
Why is taking self as a parameter for static methods better than using a weak reference for the bound method? Everything I've been looking into tells me not to do that. My PyLint checker throws up a warning when I attempt to do that.

dvarrazzo:
It is replacing a mechanism fundamentally "magic" with something less magic.

What have you read around about it not being a good idea? It's like saying
that writing a function that taking an object as parameter is a bad idea...
because that's what it is.

PyLint is full of... bad ideas. However, what's the problem with it? Just
that the name of the parameter is "self"? You can call it "obj" if you
prefer.

PR #1682, opened 2024-03-11
Add connection.get_host_addr()

exg:
In some use cases `connection.cancel()` is not usable because it may block the interpreter for a long time (as it does not release the GIL before calling `PQcancel` and does not support setting the tcp timeout for the cancel connection). An alternative approach to (asynchronously) cancel a query is to call `pg_cancel_backend` over a new connection. For this to work reliably, the connection should be established against the server IP address associated to the connection where the query is running. To this end, this PR adds a connection method to get the info using `PQhostaddr`.

### DIFF
--- a/doc/src/connection.rst
+++ b/doc/src/connection.rst
@@ -910,6 +910,20 @@ The ``connection`` class
         .. versionadded:: 2.7
 
 
+    .. index::
+        pair: Connection; Parameters
+
+    .. method:: get_host_addr()
+
+        Get the server IP address of the active connection.
+
+        .. seealso:: libpq docs for `PQhostaddr()`__ for details.
+
+            .. __: https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQHOSTADDR
+
+        .. versionadded:: 2.9.10
+
+
 .. testcode::
     :hide:
 
--- a/psycopg/connection_type.c
+++ b/psycopg/connection_type.c
@@ -947,6 +947,16 @@ exit:
 #endif
 }
 
+#define psyco_conn_get_host_addr_doc \
+"get_host_addr() -- Get the server IP address of the active connection."
+
+static PyObject *
+psyco_conn_get_host_addr(connectionObject *self, PyObject *dummy)
+{
+    EXC_IF_CONN_CLOSED(self);
+
+    return Text_FromUTF8(PQhostaddr(self->pgconn));
+}
 
 /* lobject method - allocate a new lobject */
 
@@ -1220,6 +1230,8 @@ static struct PyMethodDef connectionObject_methods[] = {
      METH_VARARGS, psyco_conn_get_parameter_status_doc},
     {"get_dsn_parameters", (PyCFunction)psyco_conn_get_dsn_parameters,
      METH_NOARGS, psyco_conn_get_dsn_parameters_doc},
+    {"get_host_addr", (PyCFunction)psyco_conn_get_host_addr,
+     METH_NOARGS, psyco_conn_get_host_addr_doc},
     {"get_backend_pid", (PyCFunction)psyco_conn_get_backend_pid,
      METH_NOARGS, psyco_conn_get_backend_pid_doc},
     {"lobject", (PyCFunction)psyco_conn_lobject,
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1826,6 +1826,9 @@ class TestConnectionInfo(ConnectingTestCase):
         self.assertEqual(d['dbname'], dbname)  # the only param we can check reliably
         self.assert_('password' not in d, d)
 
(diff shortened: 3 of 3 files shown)


dvarrazzo:
What's wrong with [`connection.info.host`](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.host)?

dvarrazzo:
I also don't see why you need the host... you need the whole set of connection parameters in order to establish a connection and to issue a `pg_cancel_backend()`.

exg:
> What's wrong with [`connection.info.host`](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.host)?

In the case where `host` is a domain name, we need the resolved IP address, as the resolution may yield a different IP address (and different database) at cancellation time.

dvarrazzo:
Ah, I see: you are after a `PQhostaddr` wrapper, not `PQhost`. So if any you should add a `info.hostaddr` method, right?

However, I am reluctant to add new interfaces to psycopg2. `PQhostaddr` was introduced only in PG 12 and I can see an entire generation of bugs arising because people might have old libpq libraries still on their system. StackOverflow is still littered of the corpses of such type of mismatches with previous optionally-compiled libpq functions.

The function you need [is already exposed by psycopg3](https://www.psycopg.org/psycopg3/docs/api/objects.html#psycopg.ConnectionInfo.hostaddr), which is implemented in a way to manage gracefully the cases in which libpq 12 wasn't available at compile time or is not available at runtime, see the [low-level PQhostaddr wrapper](https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/pq/_pq_ctypes.py#L190-L203).

So, all in all, I suggest you to move to psycopg 3.

exg:
> Ah, I see: you are after a `PQhostaddr` wrapper, not `PQhost`. So if any you should add a `info.hostaddr` method, right?

You are right, I revised the [change](https://github.com/exg/psycopg2/commits/get_host_addr/).

> However, I am reluctant to add new interfaces to psycopg2. `PQhostaddr` was introduced only in PG 12 and I can see an entire generation of bugs arising because people might have old libpq libraries still on their system. StackOverflow is still littered of the corpses of such type of mismatches with previous optionally-compiled libpq functions.

I understand your concern. However, wouldn't it be enough to check the version at runtime to avoid such issues? `PQlibVersion` was introduced in PG 9.1. The revised change includes this guard.

> The function you need [is already exposed by psycopg3](https://www.psycopg.org/psycopg3/docs/api/objects.html#psycopg.ConnectionInfo.hostaddr), which is implemented in a way to manage gracefully the cases in which libpq 12 wasn't available at compile time or is not available at runtime, see the [low-level PQhostaddr wrapper](https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/pq/_pq_ctypes.py#L190-L203).
> 
> So, all in all, I suggest you to move to psycopg 3.

That would be ideal. However, due to time constraints, I need to stick to psycopg2 for the time being. Is there any chance you could reconsider?

dvarrazzo:
> I understand your concern. However, wouldn't it be enough to check the version at runtime to avoid such issues? PQlibVersion was introduced in PG 9.1. The revised change includes this guard.

Unfortunately C is not so forgiving and problems can happen with system where pg_config belongs to a newer version than the library that will be found at runtime. Not quite DLhell, but close.

>> So, all in all, I suggest you to move to psycopg 3.

> That would be ideal. However, due to time constraints, I need to stick to psycopg2 for the time being. Is there any chance you could reconsider?

Not for free. Asking for development at a level that goes out of the lifetime of a free software project is exactly one of the ways a company can contribute to free software maintenance. 

But I have a hack for you. The [`conn.pgconn_ptr`](https://www.psycopg.org/docs/connection.html#connection.pgconn_ptr) was added for cases like this. You can use ctypes to access the libpq function and pass it this pointer.

```python
# test-hostaddr.py 

import ctypes
import ctypes.util

import psycopg2
conn = psycopg2.connect("host=localhost")

libname = ctypes.util.find_library('pq')
assert libname
libpq = ctypes.pydll.LoadLibrary(libname)
assert libpq

libpq.PQhostaddr.argtypes = [ctypes.c_void_p]
libpq.PQhostaddr.restype = ctypes.c_char_p
print("hostaddr", libpq.PQhostaddr(conn.pgconn_ptr))
```

```
$ python test-hostaddr.py 
hostaddr b'127.0.0.1'
```

Note that this only works if you use `psycopg2`, not `psycopg2-binary`. In the latter case I assume you must load the libpq library bundled with the python package. I'll leave that as an exercise (I wasted a decent amount of time figuring this out the hard way).

Issue #1692, opened 2024-04-16
Psycopg2 doesn't install on Python 3.13 alpha


junepark678:
**This is a bug tracker**
If you have a question, such has "how do you do X with Python/PostgreSQL/psycopg2" please [write to the mailing list](https://lists.postgresql.org/manage/) or [open a question](https://github.com/psycopg/psycopg2/discussions) instead.

**Before opening this ticket, please confirm that:**
- [X] I am running the latest version of pip, i.e. typing ``pip --version`` you get [this version](https://pypi.org/project/pip/).
- [X] I have read the [installation documentation](https://www.psycopg.org/docs/install.html) and the [frequently asked questions](https://www.psycopg.org/docs/faq.html)
- [X] If install failed, I typed `pg_config` on the command line and I obtained an output instead of an error.

**Please complete the following information:**
- OS: Linux (6.8.5-zen1-1-zen) on Arch Linux
- Psycopg version: 2.9.9
- Python version: 3.13.0a6+ (heads/main:6dc661bc9f)
- PostgreSQL version: 16.2
- pip version: 24.0

hroncok:
https://github.com/psycopg/psycopg2/blob/a971c11d5038333f08821ef452ce1237a22cdb0e/psycopg/utils.c#L395-L397

This function was made public and was renamed to PyInterpreterState_Get in Python 3.13.0a1+ in  https://github.com/python/cpython/pull/106321

befeleme:
And after a rename of the function, this is the test run:

```
======================================================================
FAIL: test_cidr_adapt (tests.test_ipaddress.NetworkingTestCase.test_cidr_adapt)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/psycopg2-2.9.9/tests/test_ipaddress.py", line 112, in test_cidr_adapt
    self.assertEquals(cur.fetchone()[0], '::ffff:102:300/128')
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '::ffff:1.2.3.0/128' != '::ffff:102:300/128'
- ::ffff:1.2.3.0/128
?         ^ ^ ^
+ ::ffff:102:300/128
?         ^ ^ ^


======================================================================
FAIL: test_inet_adapt (tests.test_ipaddress.NetworkingTestCase.test_inet_adapt)
----------------------------------------------------------------------
[...]

PR #1695, opened 2024-04-24
Add support for Python 3.13

befeleme:
See individual commits.
I don't have all the bits to run tests, but I used these changes to build psycopg2 in Fedora and tests with Python 3.13 passed.
It would probably be good to add 3.13 to CI, but I'm not sure about the desired version matrix, so haven't done any changes there.
Related to #1692.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,7 @@ jobs:
           - {python: "3.10", postgres: "14"}
           - {python: "3.11", postgres: "15"}
           - {python: "3.12", postgres: "16"}
+          - {python: "3.13-dev", postgres: "16"}
 
           # Opposite extremes of the supported Py/PG range, other architecture
           - {python: "3.7", postgres: "16", architecture: "x86"}
@@ -27,6 +28,7 @@ jobs:
           - {python: "3.10", postgres: "13", architecture: "x86"}
           - {python: "3.11", postgres: "11", architecture: "x86"}
           - {python: "3.12", postgres: "10", architecture: "x86"}
+          - {python: "3.13-dev", postgres: "10", architecture: "x86"}
 
     env:
       PSYCOPG2_TESTDB: postgres
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,12 @@
 Current release
 ---------------
 
+What's new in psycopg 2.9.10 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Add support for Python 3.13.
+
+
 What's new in psycopg 2.9.9
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/psycopg/utils.c
+++ b/psycopg/utils.c
@@ -392,7 +392,10 @@ psyco_set_error(PyObject *exc, cursorObject *curs, const char *msg)
 static int
 psyco_is_main_interp(void)
 {
-#if PY_VERSION_HEX >= 0x03080000
+#if PY_VERSION_HEX >= 0x030d0000
+    /* tested with Python 3.13.0a6 */
+    return PyInterpreterState_Get() == PyInterpreterState_Main();
+#elif PY_VERSION_HEX >= 0x03080000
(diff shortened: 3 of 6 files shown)


dvarrazzo:
Thank you very much for the MR!

It would actually be good to add 3.13 to the test grid.

According to the [python action docs](https://github.com/actions/setup-python), the version supported are in the [versions-manifest.json](https://github.com/actions/python-versions/blob/main/versions-manifest.json), and it seems that `3.13.0-alpha.6` could be a valid one.

Could you please try to add it to the grid? Just for Linux is fine. You can use any Postgres version you prefer: copying the same used in Python 3.11, or stretching the postgres range (you might see that there's a version missing because there are currently 6 tested Python version with 7 postgres versions, so we can close the gap and test them 1:1).

befeleme:
I added 3.13 to tox.ini and github CI config file.

dvarrazzo:
Thank you for this!

There are a few other places to touch up: the classifiers in `setup.py`

https://github.com/psycopg/psycopg2/blob/a971c11d5038333f08821ef452ce1237a22cdb0e/setup.py#L47-L61

which reminds to update `install.rst` too.

I can deal with the website once merged.

Thank you very much!

dvarrazzo:
> I added 3.13 to tox.ini and github CI config file.

The tests don't seem to pass. There is a mismatch between tox and github afaics: https://github.com/psycopg/psycopg2/actions/runs/8828621201

befeleme:
Could you rerun the tests? I redefined Python version in the config files to `3.13` & added the classifiers.

dvarrazzo:
Hmmm... there's still [a mismatch between tox and github](https://github.com/psycopg/psycopg2/actions/runs/8828621201).

I think something similar was solved in the past, for instance see this commment: 5fb1305a1420ecb884e581d84ea520e490dd5fdf

However, TBH, I am happy to drop support for tox. In the past it was useful, but now it's just a pain in the proverbial.

I think the `Install tox` step can be dropped, the `Run tests` step can be replaced by `run: make check`, the `[tox]` and `[testenv]` stanzas can be dropped from `tox.ini`.

Do you want to try with these changes?

befeleme:
Ok, another, hopefully the last attempt - I tested it on my fork and all should pass. I don't feel brave enough to reconfigure the whole CI, so I'll leave it to your consideration.

dvarrazzo:
That's ok, I appreciate your help. I'll look into doing some cleanup to the pipeline if tox gets in the way.

rouilj:
Any idea when this will be available on pypi?

I have a beta release of the Roundup Issue Tracker in a couple of weeks and a 2.4.0 release
in mid July. I would like to test the postgres database with psycopg2 and 3.13beta1 before
then in case I have to make any changes.

Thanks.

dvarrazzo:
It will be available on PyPI only after final release of 3.13.

dvarrazzo:
Merging this branch, thank you for your contribution.

We will need to set up build, and I already know that there is some bitrot in the build infrastructure.

aftermatch1320:
what's the fix for this ?
I'm still getting

ImportError: Error relocating /app/venv/lib/python3.13/site-packages/psycopg2/_psycopg.cpython-313-x86_64-linux-musl.so: _PyInterpreterState_Get: symbol not found
During handling of the above exception, another exception occurred:

dvarrazzo:
@aftermatch1320 even after we will have merged this branch, binary files for Python 3.13 will not be available until Appveyor will have provided Windows runner. This usually takes a few week after the final release of a new major version.

If you want to use psycopg on Python 3.13 on windows right now you can use psycopg 3 instead.

TiBEARius:
I'm having an issue where I try to import psycopg2 and it says that the module could not be found.  So far, I've tried uninstalling and reinstalling everything python related (including the psycopg2).  Whenever I run 'python -m pip show psycopg2' command, the latest version is clearly there in the Python site-packages folder. I've attempted hours of troubleshooting with chat gpt and haven't been able to resolve the issue. Also, python is in the advanced system settings path.  
Currently, I'm opening my file through command prompt.
What else can I do to get this to work?

Issue #1727, opened 2024-09-09
Notification is not added to notifies list when received during the commit


romank0:
- OS: I observe this on linux and macos but I can see this is os independent.
- Psycopg version: 2.9.9
- Python version: 3.11.9
- PostgreSQL version: 13.15
- pip version 24.0

## The issue

I'm using the recommended way to receive async notifications that is:

1. invoke `select.select` to wait on the data available on the connection
2. invoke `connection.poll`
3. process `connection.notifies` list. Each notification is processed in a transaction

Basically the code is almost the copy and paste from the documentation except that I do not use autocommit because I manually manage transactions:
```
curs = conn.cursor()
curs.execute("LISTEN test;")
conn.commit()

while True:
    if select.select([conn],[],[],1) == ([],[],[]):
        print("Timeout")
    else:
        conn.poll()
        while conn.notifies:
            notify = conn.notifies.pop(0)
            print("Got NOTIFY:", notify.pid, notify.channel, notify.payload)
            # some processing that updates the DB based on the notify content
            conn.commit()
```

The problem is that sometimes (specifically when the notification is received during the `commit`) this code gets into the state that:
- there is no data to read from the socket (as `select.select` shows)
- but if I invoke `poll` on the connection new notifications will be added to `notifies` list

Because of this the code loops on the `select.select` and some notifications processing is delayed until the new notification comes in later.

# My interpretation of what happens

I've tracked this down to the problem in the `pq_commit` and/or `pq_execute_command_locked`. Namely, according to pqlib [documentation](https://www.postgresql.org/docs/16/libpq-notify.html):

> You should, however, remember to check PQnotifies after each [PQgetResult](https://www.postgresql.org/docs/16/libpq-async.html#LIBPQ-PQGETRESULT) or [PQexec](https://www.postgresql.org/docs/16/libpq-exec.html#LIBPQ-PQEXEC), to see if any notifications came in during the processing of the command.

`pq_execute_command_locked` does invoke `PQexec` and that according to my understanding reads up the data from the socket but then `PQnotifies` are not checked in `pq_commit`. So the followup invocation of `select` shows there is not more data to read and there is no progress in the listen loop.

Most probably this is not the only place this happens. For example in `pq_get_result_async` I can see that notifications are checked _before_ executing `PQgetResult` but according to pqlib docs above that should happen after.

### The script reproducing the issue.

```
import threading
import select
import psycopg2
import psycopg2.extensions
import time

[...]

PR #1728, opened 2024-09-10
Adds notifies processing during commit

romank0:
This fixes the issue that asynchronous notifications are not fetched during the commit https://github.com/psycopg/psycopg2/issues/1727.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,12 +1,16 @@
-Current release
+Future releases
 ---------------
 
 What's new in psycopg 2.9.10 (unreleased)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Add support for Python 3.13.
+- Receive notifications on commit (:ticket:`#1728`).
 - Drop support for Python 3.7.
 
+
+Current release
+---------------
 
 What's new in psycopg 2.9.9
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/psycopg/connection_int.c
+++ b/psycopg/connection_int.c
@@ -1344,6 +1344,11 @@ conn_set_session(connectionObject *self, int autocommit,
         }
     }
 
+    Py_BLOCK_THREADS;
+    conn_notifies_process(self);
+    conn_notice_process(self);
+    Py_UNBLOCK_THREADS;
+
     if (autocommit != SRV_STATE_UNCHANGED) {
         self->autocommit = autocommit;
     }
@@ -1407,6 +1412,11 @@ conn_set_client_encoding(connectionObject *self, const char *pgenc)
     if ((res = pq_set_guc_locked(self, "client_encoding", clean_enc, &_save))) {
         goto endlock;
     }
+
+    Py_BLOCK_THREADS;
+    conn_notifies_process(self);
+    conn_notice_process(self);
+    Py_UNBLOCK_THREADS;
 
 endlock:
     pthread_mutex_unlock(&self->lock);
--- a/psycopg/pqpath.c
+++ b/psycopg/pqpath.c
@@ -412,6 +412,7 @@ pq_commit(connectionObject *conn)
     }
 
     Py_BLOCK_THREADS;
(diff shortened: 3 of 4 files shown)


romank0:
By some reason the test fails in CI 😕  It might be some difference in notifications delivery in a different version of postgres.

dvarrazzo:
Hello,

your change makes sense.

I wonder if we shouldn't maintain the symmetry and add the `conn_notifies_process()` call after ROLLBACK too. From a semantic point of view it should do nothing because notifications generated in a transaction are not sent in case of transaction rollback, but from a protocol point of view some notification might be in that message.

About the test failing, I wonder is if there isn't any race condition that gets triggered when the server is remote. Where do you see the failing test?

romank0:
> From a semantic point of view it should do nothing because notifications generated in a transaction are not sent in case of transaction rollback,

The notifications that are generated in the _current_ transaction will not be delivered by postgres but notifications generated by other committed transaction may be delivered in any interaction with the server including the rollback.

I think I can add the similar processing for the rollback that will by no means be the complete fix. This fix is very limited in its scope namely this fixes only the commit.
The proper fix I belive should follow the guidelines from pqlib:

> You should, however, remember to check PQnotifies after each PQgetResult or PQexec, to see if any notifications came in during the processing of the command

This means that the notifies checking should happen at every call to these two functions. I'm not however comfortable with doing such massive change without proper understanding how the code works w.r.t. locking and C<->python interaction. Specifically I don't understand if it possible to move the notifies checking into `pq_execute_command_locked` which is one of the places where `PQexec` is called.

> Where do you see the failing test

image

On this check there are multiple steps. Some of them fail due to some configuration reason as in test are not run at all:
```
Running Install scripts
%PYEXE% scripts\build\appveyor.py install
The system cannot find the path specified.
Command exited with code 1
```

But some steps fail because of the new test that I've added:
```
======================================================================
FAIL: test_notifies_received_on_commit (tests.test_notify.NotifiesTests)
----------------------------------------------------------------------
  File "C:\Project\tests\testutils.py", line 459, in skip_if_crdb__
    return f(self, *args, **kwargs)
  File "C:\Project\tests\testutils.py", line 529, in slow_
    return f(self)
  File "C:\Project\tests\test_notify.py", line 137, in test_notifies_received_on_commit
    self.assertEqual(1, len(self.conn.notifies))
AssertionError: 1 != 0
----------------------------------------------------------------------
Ran 743 tests in 161.149s
FAILED (failures=1, skipped=33)
  File "scripts\build\appveyor.py", line 846, in <module>
    sys.exit(main())
  File "scripts\build\appveyor.py", line 39, in main
    cmd()
  File "scripts\build\appveyor.py", line 386, in step_test_script
    run_test_suite()
  File "scripts\build\appveyor.py", line 422, in run_test_suite
    run_python(args)
  File "scripts\build\appveyor.py", line 609, in run_python
    return run_command([opt.py_exe] + args, **kwargs)
  File "scripts\build\appveyor.py", line 593, in run_command
    sp.check_call(cmdline, **kwargs)
  File "C:\Python36\lib\subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['C:\\Python38\\python.exe', '-c', "import tests; tests.unittest.main(defaultTest='tests.test_suite')", '--verbose']' returned non-zero exit status 1.
```

dvarrazzo:
Ah, on windows, I see. That's strange: this doesn't seem a platform-dependent change. I would be tempted to just skip the test on window. Of course the tests whose configuration fail are not your fault: the appveyor pipeline hasn't been exercised in a while probably.

romank0:
I've added notifications processing to all places where `PQexec` is invoked and disabled tests on windows. I would like to parametrize tests to avoid repetitive code but I am not sure about bringing additional dependencies.

dvarrazzo:
We massaged a bit Appveyor in order to make it build Python 13 package and as a result tests now pass. I think we can merge this. Thank you very much!

Issue #1724, opened 2024-08-24
Incorrect Decimal('Infinity') adaptation


hordiienko-v:
- OS: Ubuntu
- Psycopg version: 2.9.9
- Python version: 3.9
- PostgreSQL version: 16.2
- pip version: 21.2.4

Since v14, PostgreSQL NUMERIC type supports `Infinity` value. But for now, psycopg2 adapts `Infinity` as `'NaN'::numeric`. This can be avoided by creating custom type adapter that wraps decimal value in quotes, but still, adapting to `NaN` seems like outdated logic.

```python
from decimal import Decimal

from psycopg2 import connect
from psycopg2._psycopg import connection, cursor

if __name__ == "__main__":
    conn: connection
    curr: cursor

    with connect("dbname=test_db user=test_user password=test_user") as conn:
        with conn.cursor() as curr:
            q = curr.mogrify(
                "INSERT INTO test VALUES (%s), (%s)",
                [Decimal("1"), Decimal("Infinity")],
            )
            print(q)
            # b"INSERT INTO test VALUES (1), ('NaN'::numeric)"
```

dvarrazzo:
Thank you for the report. The solution is probably little enough invasive that we can provide it in the next bugfix release.

Psycopg 3 already does the right thing:

```python
>>> import psycopg.sql
>>> from decimal import Decimal
>>> psycopg.sql.quote(Decimal("Infinity"))
"'Infinity'::numeric"
```

edgarrmondragon:
> Thank you for the report. The solution is probably little enough invasive that we can provide it in the next bugfix release.
> 
> Psycopg 3 already does the right thing:
> 
> ```python
> >>> import psycopg.sql
> >>> from decimal import Decimal
> >>> psycopg.sql.quote(Decimal("Infinity"))
> "'Infinity'::numeric"
> ```

Got a fix for that: https://github.com/psycopg/psycopg2/pull/1735

dvarrazzo:
I am reconsidering the opportunity of changing this behaviour. Yes, converting inf -> nan is not correct but otoh I don't know if there are applications out there assuming it...

PR #1729, opened 2024-09-16
Build Python 3.13 wheels, drop support for Python 3.7

edgarrmondragon:
I found it quite difficult to continue building 3.7 wheels for macos. Maybe someone knows better.

See https://github.com/psycopg/psycopg2/issues/1692.


doc/src/install.rst
```diff
@@ -131,7 +131,7 @@ The current `!psycopg2` implementation supports:
 ..
     NOTE: keep consistent with setup.py and the /features/ page.
 
-- Python versions from 3.7 to 3.12
+- Python versions from 3.8 to 3.12
```
dvarrazzo, line 134: This can go to 3.13

dvarrazzo:
Trying to [build the packages](https://github.com/psycopg/psycopg2/actions/runs/11232596331)... :eyes:

doc/src/install.rst
dvarrazzo: Done it myself

dvarrazzo:
Looking into the build failure on appveyor. It seems like the database used for tests is too old, trying to use a more recent one.

edgarrmondragon:
@dvarrazzo are there any more blockers?

dvarrazzo:
Yesterday I managed to run the test, but not the packaging. Trying that now.

dvarrazzo:
I managed to build packages on Windows too. We need to drop Python 3.8 but it doesn't seem a problem.

There are a couple of little bugs I'd like to fix:

- #1728 to re-test and merge
- #1724 to fix, but it should be easy

So I'd like to get those in and make a release.

PR #1735, opened 2024-10-11
Fix `Decimal('Infinity')` adaptation

edgarrmondragon:
Closes https://github.com/psycopg/psycopg2/issues/1724

### DIFF
--- a/psycopg/adapter_pdecimal.c
+++ b/psycopg/adapter_pdecimal.c
@@ -47,8 +47,21 @@ pdecimal_getquoted(pdecimalObject *self, PyObject *args)
         }
         goto output;
     }
-    else if (check) {
+
+    check = PyObject_CallMethod(self->wrapped, "is_nan", NULL);
+    if (check == Py_True) {
         res = Bytes_FromString("'NaN'::numeric");
+        goto end;
+    }
+
+    /* If the decimal is not finite and not NaN, it must be infinity,
+     * so all that is left is to check if it is positive or negative. */
+    check = PyObject_CallMethod(self->wrapped, "is_signed", NULL);
+    if (check == Py_True) {
+        res = Bytes_FromString("'-Infinity'::numeric");
+        goto end;
+    } else {
+        res = Bytes_FromString("'Infinity'::numeric");
         goto end;
     }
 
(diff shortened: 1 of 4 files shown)


tests/test_types_basic.py
```diff
@@ -74,12 +74,15 @@ def testDecimal(self):
         self.failUnless(str(s) == "NaN", "wrong decimal quoting: " + str(s))
         self.failUnless(type(s) == decimal.Decimal,
                         "wrong decimal conversion: " + repr(s))
+
+    @testutils.skip_before_postgres(14, 0)
```
edgarrmondragon, line 78: `SELECT 'Infinity'::numeric AS foo` causes

```
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type numeric: "Infinity"
```

in Postgres 12 and 13.

edgarrmondragon: Oh, now that I think about it more we should probably rather check the server version when adapting `Decimal('Infinity')`, and fall back to `NaN'::numeric` on Postgres < 14.

dvarrazzo: Uhm... Thinking a bit harder abut it: is it even a good idea to change the way inf is converted? Yes, converting it to nan wasn't the right thing to do, but maybe there is code that assumes that behaviour.

I am reconsidering the opportunity of making this change.

Note: psycopg 3 doesn't check for version and doesn't convert inf -> nan. Trying to send an inf results in a server error, which is the right thing to do.

psycopg/adapter_pdecimal.c
```diff
@@ -47,11 +47,24 @@ pdecimal_getquoted(pdecimalObject *self, PyObject *args)
...
+    check = PyObject_CallMethod(self->wrapped, "is_signed", NULL);
+    if (check == Py_True) {
+        res = Bytes_FromString("'-Infinity'::numeric");
+        goto end;
+    } else {
+        res = Bytes_FromString("'Infinity'::numeric");
+        goto end;
+    }
+
     /* is_finite() was introduced 2.5.1 < somewhere <= 2.5.4.
```
dvarrazzo, line 68: Wow, I'm not sure this code is still used :smile:

dvarrazzo (changes requested):

tests/test_types_basic.py
edgarrmondragon: I agree. Either option (converting it or failing) is a potential breaking change. Happy to close this if you think we err on preserving current behavior.

dvarrazzo:
Thank you very much for this MR, but as discussed it's probably not a good idea to do this in psycopg2.

PR #1755, opened 2024-11-22
ci(macos): Avoid linking against Homebrew

### LINKED ISSUE #1753: High macOS deployment targets for psycopg2-binary
**This is a bug tracker**
If you have a question, such has "how do you do X with Python/PostgreSQL/psycopg2" please [write to the mailing list](https://lists.postgresql.org/manage/) or [open a question](https://github.com/psycopg/psycopg2/discussions) instead.

**Before opening this ticket, please confirm that:**
- [x] I am running the latest version of pip, i.e. typing ``pip --version`` you get [this version](https://pypi.org/project/pip/).
- [x] I have read the [installation documentation](https://www.psycopg.org/docs/install.html) and the [frequently asked questions](https://www.psycopg.org/docs/faq.html)
- [ ] If install failed, I typed `pg_config` on the command line and I obtained an output instead of an error.

**Please complete the following information:**
- OS: macOS 13 arm64 
- Psycopg version: 2.9.10
- Python version: 3.12.7
- PostgreSQL version:
- pip version: 24.3.1

The `MACOSX_DEPLOYMENT_TARGET` sets the minimum version of macOS that the `psycopg2-binary` wheels will install on so the introduction of [setting MACOSX_DEPLOYMENT_TARGET to the CI runner version](https://github.com/psycopg/psycopg2/commit/a59079a4f2c4af12224a385eff4397954b647293#diff-5bcf795eb775d7afa43ccd6534c336b59bc441c814c20f4b30603fa7d50fb03cR171) blocks installs on anything older than the CI runner. As of `2.9.10`, `pip install psycopg2-binary` on a macOS <14.0 arm64 or <12.0 x86_64 machine will ignore the wheels, pull the sdist, probably fail to build from source and give the misleading advice to try `pip install psycopg2-binary` instead.

Judging by the lack of *drop support for macOS 13* entries in the changelog and macOS 13 still having another year before its EOL date, I'm guessing that wasn't intentional?

Incidentally, why publish sdists for the `psycopg2-binary` package?

bwoodsend:
Homebrew binaries are always compiled for exactly the version they're installed on making them very unportable. When a wheel is "repaired" by cibuildwheel, delocate-wheel pulls in _psycopg's dependencies (libpq.dylib, libssl.dylib and libcrypto.dylib) which, on a GitHub Actions macOS 14 runner, are provided by Homebrew and are therefore only macOS >= 14 compatible. The resultant wheel is therefore incompatible with all but the latest macOS versions.

Build all dependencies from source so that we can set the deployment target to something sensible. Fixes #1753.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -132,29 +132,15 @@ jobs:
   # }}}
 
   macos:  # {{{
-    runs-on: macos-${{ matrix.macver }}
+    runs-on: macos-latest
     if: true
 
     strategy:
       fail-fast: false
       matrix:
         # These archs require an Apple M1 runner: [arm64, universal2]
-        arch: [x86_64]
+        arch: [x86_64, arm64]
         pyver: [cp39, cp310, cp311, cp312, cp313]
-        macver: ["13"]
-        include:
-          - arch: arm64
-            pyver: cp310
-            macver: "14"
-          - arch: arm64
-            pyver: cp311
-            macver: "14"
-          - arch: arm64
-            pyver: cp312
-            macver: "14"
-          - arch: arm64
-            pyver: cp313
-            macver: "14"
 
     steps:
       - name: Checkout repos
@@ -171,6 +157,7 @@ jobs:
         env:
           CIBW_BUILD: ${{matrix.pyver}}-macosx_${{matrix.arch}}
           CIBW_ARCHS_MACOS: ${{matrix.arch}}
+          MACOSX_ARCHITECTURE: ${{matrix.arch}}
           CIBW_BEFORE_ALL_MACOS: ./scripts/build/wheel_macos_before_all.sh
           CIBW_TEST_COMMAND: >-
             export PYTHONPATH={project} &&
@@ -184,7 +171,7 @@ jobs:
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: macos-${{matrix.pyver}}-macos-${{matrix.macver}}_${{matrix.arch}}
+          name: macos-${{matrix.pyver}}-macos-${{matrix.arch}}
           path: ./wheelhouse/*.whl
 
   # }}}
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -5,6 +5,8 @@ on:
 
 env:
   PIP_BREAK_SYSTEM_PACKAGES: "1"
+  LIBPQ_VERSION: "16.0"
+  OPENSSL_VERSION: "1.1.1w"
 
 jobs:
   sdist:  # {{{
@@ -58,10 +60,6 @@ jobs:
 
   linux:  # {{{
     if: true
-
-    env:
-      LIBPQ_VERSION: "16.0"
-      OPENSSL_VERSION: "1.1.1w"
 
     strategy:
       fail-fast: false
@@ -162,6 +160,12 @@ jobs:
       - name: Checkout repos
         uses: actions/checkout@v4
 
+      - name: Build dependencies
+        run: ./scripts/build/build_libpq.sh
+
+      - name: Show dependency tree
+        run: otool -L /tmp/libpq.build/lib/*.dylib
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.22.0
         env:
@@ -172,12 +176,11 @@ jobs:
             export PYTHONPATH={project} &&
             python -c "import tests; tests.unittest.main(defaultTest='tests.test_suite')"
           CIBW_ENVIRONMENT: >-
-            MACOSX_DEPLOYMENT_TARGET=${{ matrix.macver }}.0
             PG_VERSION=16
             PACKAGE_NAME=psycopg2-binary
             PSYCOPG2_TESTDB=postgres
             PSYCOPG2_TEST_FAST=1
-            PATH="/usr/local/opt/postgresql@${PG_VERSION}/bin:$PATH"
+            PATH="/tmp/libpq.build/bin:$PATH"
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -160,6 +160,12 @@ jobs:
       - name: Checkout repos
         uses: actions/checkout@v4
 
+      - name: Cache libpq build
+        uses: actions/cache@v4
+        with:
+          path: /tmp/libpq.build
+          key: libpq-${{ env.LIBPQ_VERSION }}-macos-${{ matrix.arch }}
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.22.0
         env:
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -179,7 +179,6 @@ jobs:
             PG_VERSION=16
             PACKAGE_NAME=psycopg2-binary
             PSYCOPG2_TESTDB=postgres
-            PSYCOPG2_TEST_FAST=1
             PATH="/tmp/libpq.build/bin:$PATH"
 
       - name: Upload artifacts
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -160,12 +160,6 @@ jobs:
       - name: Checkout repos
         uses: actions/checkout@v4
 
-      - name: Build dependencies
-        run: ./scripts/build/build_libpq.sh
-
-      - name: Show dependency tree
-        run: otool -L /tmp/libpq.build/lib/*.dylib
-
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.22.0
         env:
--- a/scripts/build/build_libpq.sh
+++ b/scripts/build/build_libpq.sh
@@ -68,9 +68,12 @@ case "$ID" in
         # it, force use of system curl.
         brew uninstall --force --ignore-dependencies openssl gettext
         curl="/usr/bin/curl"
-        # The deployment target should be <= to that of the oldest supported Python version.
+        if [ -z "$MACOSX_ARCHITECTURE" ]; then
+            MACOSX_ARCHITECTURE="$(uname -m)"
+        fi
+        # Set the deployment target to be <= to that of the oldest supported Python version.
         # e.g. https://www.python.org/downloads/release/python-380/
-        if [ "$(uname -m)" == "x86_64" ]; then
+        if [ "$MACOSX_ARCHITECTURE" == "x86_64" ]; then
             export MACOSX_DEPLOYMENT_TARGET=10.9
         else
             export MACOSX_DEPLOYMENT_TARGET=11.0
@@ -83,6 +86,22 @@ case "$ID" in
         ;;
 esac
 
+
+if [ "$ID" == "macos" ]; then
+    make_configure_standard_flags=( \
+        --prefix=${LIBPQ_BUILD_PREFIX} \
+        "CPPFLAGS=-I${LIBPQ_BUILD_PREFIX}/include/ -arch $MACOSX_ARCHITECTURE" \
+        "LDFLAGS=-L${LIBPQ_BUILD_PREFIX}/lib -arch $MACOSX_ARCHITECTURE" \
+    )
+else
+    make_configure_standard_flags=( \
+        --prefix=${LIBPQ_BUILD_PREFIX} \
+        CPPFLAGS=-I${LIBPQ_BUILD_PREFIX}/include/ \
+        LDFLAGS=-L${LIBPQ_BUILD_PREFIX}/lib \
+    )
+fi
+
+
 if [ "$ID" == "centos" ] || [ "$ID" == "macos" ]; then
 
     # Build openssl if needed
@@ -94,8 +113,14 @@ if [ "$ID" == "centos" ] || [ "$ID" == "macos" ]; then
 
         cd "${openssl_dir}"
 
-        ./config --prefix=${LIBPQ_BUILD_PREFIX} --openssldir=${LIBPQ_BUILD_PREFIX} \
-            zlib -fPIC shared
+        options=(--prefix=${LIBPQ_BUILD_PREFIX} --openssldir=${LIBPQ_BUILD_PREFIX} \
+            zlib -fPIC shared)
+        if [ -z "$MACOSX_ARCHITECTURE" ]; then
+            ./config $options
+        else
+            ./configure "darwin64-$MACOSX_ARCHITECTURE-cc" $options
+        fi
+
         make depend
         make
     else
@@ -119,9 +144,7 @@ if [ "$ID" == "macos" ]; then
             | tar xzf -
 
         cd "${krb5_dir}"
-
-        ./configure --prefix=${LIBPQ_BUILD_PREFIX} \
-            CPPFLAGS=-I${LIBPQ_BUILD_PREFIX}/include/ LDFLAGS=-L${LIBPQ_BUILD_PREFIX}/lib
+        ./configure "${make_configure_standard_flags[@]}"
         make
     else
         cd "${krb5_dir}"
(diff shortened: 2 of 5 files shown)


bwoodsend:
I've temporarily disabled all the other bits of CI just to make verifying quicker. I'll revert all that if/when we get closer to merging. I wanted to gauge whether you think this is worth the trouble before finishing it off? I can't imagine you'd be particularly comfortable with having a non-trivial build step that you can't run locally.

dvarrazzo:
Yes, this seems useful, and it's what we do for Linux: see the `scripts/build/build_libpq.sh`. Maybe it would be worth augmenting that script with macOS support? For sure it's better to edit complex bash in a `.sh` file than interspersed in yaml.

If you complete this operation it would be useful to port it to psycopg 3 too (which has a script pretty much identical).

Hint: you can use `if: false` to disable a job, which is less invasive then chopping everything away (the reason why there are the `if: true` in the workflow file is to remember that).

.github/workflows/packages.yml
```diff
@@ -158,6 +30,37 @@ jobs:
...
+      - name: Build postgresql from source
+        run: |
+          curl --fail-with-body -L https://ftp.postgresql.org/pub/source/v16.6/postgresql-16.6.tar.bz2 | tar x
+          brew uninstall openssl --ignore-dependencies
+
+          export MACOSX_DEPLOYMENT_TARGET=${{ env.deployment_target }}
+          export LDFLAGS="-L$fake_prefix/lib"
+          export CPPFLAGS="-I$fake_prefix/include"
+          cd postgresql-16.6
+          ./configure --without-icu --prefix="$fake_prefix" --with-ssl=openssl
```
dvarrazzo, line 57: No gssapi/ldap support?

dvarrazzo:
What is, on macOS, the default socket dir? In other words:
```
$ psql "host='' port=9999"
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.9999" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
```
what path is searched for in macOS? If it's not `/tmp` then it's worth patching the libpq before building, see

https://github.com/psycopg/psycopg2/blob/e83754a4146d01a9bbf0b8cbf0dfd085ce9394f7/scripts/build/build_libpq.sh#L148-L152

bwoodsend:
> see the scripts/build/build_libpq.sh. Maybe it would be worth augmenting that script with macOS support? For sure it's better to edit complex bash in a .sh file than interspersed in yaml.

Thanks for pointing that out. Yes, adding macOS support to that script sounds better than doing it all over again in YAML.

> What is, on macOS, the default socket dir?

I wasn't aware that there was such thing as a socket dir but does this answer your question?

```fish
> psql "host='' port=9999"
psql: error: connection to server on socket "/tmp/.s.PGSQL.9999" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
```
(That's homebrew's `psql` – not a source built one in case it matters.)

.github/workflows/packages.yml
bwoodsend: Ah, I was going to ask about which `--with-xxx` features to use but you beat me to it. I see `build_libpq.sh` makes that decision for me anyway.

bwoodsend:
Do you mind if I bump the OpenLDAP version? I need [this fix](https://git.openldap.org/openldap/openldap/-/commit/9b9434ee4add63f115ff9a855ca1afc154b6e28e#09be8533ff0a6ee5d577f971145ed449399fcda4_16371_16376) or the check for POSIX regex throws a false negative. It's probably `sed`-able if not.

.github/workflows/packages.yml
dvarrazzo: Unfortunately that also implies building these other libraries. Not doing so would break certain features. I don't know how many people rely on them, but I don't feel they are features that we can drop without being unpunished.

TBH, for some reason, GSSAPI causes a lot of problems to macOS users because in some configuration it causes slowdowns on connection, but still it doesn't seem a good case to decide that nobody in the world requires it.

dvarrazzo:
> I wasn't aware that there was such thing as a socket dir but does this answer your question?

Yes it does. It seems that the default socket dir is `/tmp/`, which is the default one in Postgres source, so we don't need patching.

It is useful that it is the brew one. If it wasn't, a program configured and running correctly so far would stop working if we switch to a libpq with a different default. See [past issues](https://github.com/psycopg/psycopg2/issues?q=DEFAULT_PGSOCKET_DIR) for example.


> Do you mind if I bump the OpenLDAP version?

Absolutely not, go ahead :slightly_smiling_face:

dvarrazzo:
Hello @bwoodsend! How is it going with this work? Is it ready, or are you stuck somewhere, or you don't think you want to finish the MR?

Thank you very much!

bwoodsend:
Sorry, I'm still here. I ground to a halt over some [m4 issue](https://github.com/bwoodsend/psycopg2/actions/runs/11999181989/job/33446811076#step:3:12179) running `cyrus-sasl`'s `autoreconf` which I couldn't reproduce locally (despite also using macOS 14 M1). I suspect there's something from homebrew on the GitHub runner shadowing the xcode version that works somewhere. When I get a quite moment where I'm more than half awake, I'll [tmate](https://github.com/mxschmitt/action-tmate) my way in and see if I can figure it out.

bwoodsend:
Yay! [Got there eventually](https://github.com/bwoodsend/psycopg2/actions/runs/12216481981). I'll clean up my mess in a bit and move this out of draft.

> I suspect there's something from homebrew on the GitHub runner shadowing the xcode version that works somewhere.

Sure enough, `brew install libtool` was the magic cure. I never really appreciated before how evil these xcode/brew tools and brew/system libraries dualities were for issue debugging and reproducibility.

The `macos-12` GitHub runners have just had the plug pulled on them. I'll bump us onto `macos-13` but I'm rather uncomfortable that this time next year, that'll be gone too and we'll have no more `x86_64` runners. At that point I suspect we'll be forced to cross compile. I've given it a brief poke locally and it looks like each component insists on doing it in their own way. So not quick and not fun...

scripts/build/build_libpq.sh
```diff
@@ -12,30 +12,69 @@ postgres_version="${LIBPQ_VERSION}"
...
         apk add --no-cache zlib-dev krb5-dev linux-pam-dev openldap-dev openssl-dev
+        curl="$(which curl)"
+        ;;
+
+    macos)
+        brew install automake m4 libtool
+        # If available, libpq seemingly insists on linking against homebrew's
+        # openssl no matter what so remove it. Since homebrew's curl depends on
+        # it, force use of system curl.
+        brew uninstall --force --ignore-dependencies openssl gettext
```
bwoodsend, line 69: Don't suppose you know a better way to force which copy of a library to link against? The build system seems to add homebrew to its search path even if I `env -i` away all the brew related environment variables.

bwoodsend:
Up to date verification run: https://github.com/bwoodsend/psycopg2/actions/runs/12291165704

dvarrazzo:
Hello @bwoodsend 

I have worked a bit on this branch, mostly adding libpq build caching and moving the libpq build from the CI workflow to the BEFORE script, consistently with the linux workflow.

A complete package build with all the version run [here](https://github.com/psycopg/psycopg2/actions/runs/12616000339/job/35156745700).

This looks ok to merge for me. Ok for you too?

After this is safe I will also port it to psycopg3, which has the same type of problem.

Thank you very much for your contribution!

bwoodsend:
Do you want to cherry pick https://github.com/bwoodsend/psycopg2/commit/ede7cd2cef12176606eaffd38284315b98068199 in or we can save it for later? I wrote it a while ago but never got around to pushing it (curse my tiny attention span).

I'm happy for this to be merged now either way.

scripts/build/build_libpq.sh
```diff
@@ -12,30 +12,69 @@ postgres_version="${LIBPQ_VERSION}"
...
         apk upgrade
         apk add --no-cache zlib-dev krb5-dev linux-pam-dev openldap-dev openssl-dev
+        curl="$(which curl)"
+        ;;
+
+    macos)
+        brew install automake m4 libtool
+        # If available, libpq seemingly insists on linking against homebrew's
+        # openssl no matter what so remove it. Since homebrew's curl depends on
+        # it, force use of system curl.
```
dvarrazzo, line 68: Question: why don't we remove homebrew curl too and just leave the system one?

scripts/build/build_libpq.sh
dvarrazzo: Maybe with `-L/-l` options in the compiler? Not sure.

scripts/build/build_libpq.sh
```diff
@@ -70,7 +109,55 @@ if [ "$ID" == "centos" ]; then
...
+    make install
+    cd ../..
+
+fi
+
+
+if [ "$ID" == "macos" ]; then
+
+    # Build gettext if needed
+    gettext_dir="gettext-${gettext_version}"
```
dvarrazzo, line 139: Which library requires gettext? Maybe we can just avoid it using `--without-gettext` or similar in `configure`.

dvarrazzo:
> Do you want to cherry pick https://github.com/bwoodsend/psycopg2/commit/ede7cd2cef12176606eaffd38284315b98068199 in or we can save it for later?

If it works, it seems good. I remember that with cross-compiling on macOS you could build but not run/test what built. Is that only a concern for cross-compiling ARM on X86?

scripts/build/build_libpq.sh
dvarrazzo: Hmm... thinking about it, the libpq does use localization. Better to have this then I guess.

dvarrazzo: Still, I [run a test](https://github.com/psycopg/psycopg2/actions/runs/12622064372/job/35169563599) and it seems that gettext is not needed, at least on x86. @bwoodsend maybe this was needed in a different test?

scripts/build/build_libpq.sh
```diff
@@ -70,7 +109,55 @@ if [ "$ID" == "centos" ]; then
...
 
 
-if [ "$ID" == "centos" ]; then
+if [ "$ID" == "macos" ]; then
+
+    # Build kerberos if needed
+    krb5_dir="krb5-${krb5_version}/src"
+    if [ ! -d "${krb5_dir}" ]; then
+        "$curl" -sL \
+            curl -sL "https://kerberos.org/dist/krb5/$(echo 1.21.3 | grep -oE '\d+\.\d+')/krb5-${krb5_version}.tar.gz" \
```
dvarrazzo, line 118: There are two errors here: curl is repeated twice, the version number stripping doesn't use the `krb5_version`. Note that you can strip a suffix using:
```
$ echo ${krb5_version%.*}
1.21
```

scripts/build/build_libpq.sh
bwoodsend: Good point. I'll give that a go.

dvarrazzo: I'm already testing it, no worries :)

dvarrazzo: Looks like it's working ok. https://github.com/psycopg/psycopg2/actions/runs/12622307173

scripts/build/wheel_macos_before_all.sh
```diff
@@ -11,6 +11,12 @@ set -x
 dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 prjdir="$( cd "${dir}/../.." && pwd )"
 
+# Build dependency libraries
+"${prjdir}/scripts/build/build_libpq.sh"
+
+# Show dependency tree
+# otool -L /tmp/libpq.build/lib/*.dylib
```
bwoodsend, line 18: Did you mean to comment the `otool` line out?

dvarrazzo: Yes, I thought it was just a debug statement. Do we need its output in CI?

scripts/build/build_libpq.sh
bwoodsend: Well that's weird. It's providing `libintl` but the `otool -L` output says nothings linked against that library so I guess it is redundant. Can't think why I'd have thought that it was a dependency.

bwoodsend:
> > Do you want to cherry pick [bwoodsend@ede7cd2](https://github.com/bwoodsend/psycopg2/commit/ede7cd2cef12176606eaffd38284315b98068199) in or we can save it for later?
> 
> If it works, it seems good. I remember that with cross-compiling on macOS you could build but not run/test what built. Is that only a concern for cross-compiling ARM on X86?
> 
> Edit: answering to myself: looking [here](https://cibuildwheel.pypa.io/en/stable/faq/#cross-compiling) it seems that x86 code can be tested on arm.

Yeah, you can run `x86_64` binaries on `arm64` using Rosetta emulation but not the other way around. I think that's also scheduled for removal in macOS 16 but we can make the most of it whilst it's still there.

scripts/build/wheel_macos_before_all.sh
bwoodsend: It is just debug but it's quite nice to see exactly why some homebrew package got sucked in every time cibuildwheel gives you that rather unhelpful error about raising deployment targets. It's also handy for answering your `gettext` question.

It's up to you.

dvarrazzo: Fair. Let me enable it back :slightly_smiling_face:

scripts/build/build_libpq.sh
bwoodsend: I'm guessing you've resolved this one because you're mopping it up yourself along with your `brew uninstall curl` change?

dvarrazzo: Yes, fixed

scripts/build/build_libpq.sh
dvarrazzo: Dropped, and it seems to work well on both architectures. Also re-enabled otool here. https://github.com/psycopg/psycopg2/actions/runs/12622655233

scripts/build/build_libpq.sh
```diff
@@ -38,39 +56,101 @@ case "$ID" in
         apk add --no-cache zlib-dev krb5-dev linux-pam-dev openldap-dev openssl-dev
         ;;
 
+    macos)
+        brew install automake m4 libtool
+        # If available, libpq seemingly insists on linking against homebrew's
+        # openssl no matter what so remove it. Since homebrew's curl depends on
+        # it, force use of system curl.
+        brew uninstall --force --ignore-dependencies openssl gettext curl
+        if [ -z "$MACOSX_ARCHITECTURE" ]; then
```
bwoodsend, line 65: I missed that the `-z "$variable"` pattern breaks with bash's *error on undefined variable* mode on. i.e. I've broken non-cross compile mode.

```suggestion
        if [ -z "${MACOSX_ARCHITECTURE-}" ]; then
```
Likewise with the `openssl` check on line 133.

Shall I `commit --amend` the fix into the offending commit. I'm slightly wary of doing a force push whilst we're both pushing to the same branch.

dvarrazzo:
Thank you for the in-depth co-review, it seems a good improvement to only depend on the ARM runners.

Building packages in [this job](https://github.com/psycopg/psycopg2/actions/runs/12622822097). If everything works I guess we can merge this.

scripts/build/build_libpq.sh
bwoodsend: Incidentally, are you happy with the name `MACOSX_ARCHITECTURE`? It's sort of the standard variable but not even standard enough that any of these packages' build systems supported it so we could go for something more memorable?

dvarrazzo: I actually thought that it was a standard variable, re-defined for an excess of prudence, but yes, it would break with `-u`. Will fix it (usually I use the `:-` suffix).

bwoodsend: Thanks.

> usually I use the :- suffix

One day I won't have to look it up every time I need to remember what all the variable expansion rules are...

dvarrazzo: re-running [a new job](https://github.com/psycopg/psycopg2/actions/runs/12622882792) after clearing the runners' caches.

bwoodsend:
That's everything I can think of. Ready for the merge button when you are...

dvarrazzo:
Thank you very much! Merged, now poriting to psycopg 3 in https://github.com/psycopg/psycopg/pull/986.

PR #1776, opened 2025-03-05
chore: bundle libpq 17

ngnpope:
Bundle libpq 17 in the binary wheels.

I've had reports of issues using psycopg2 v2.9.10 with PostgreSQL v17 on the server side:

```
ERROR:  column d.daticulocale does not exist
LINE 6:        d.daticulocale as "ICU Locale",
               ^
HINT:  Perhaps you meant to reference the column "d.datlocale".
```

The column was renamed. See https://github.com/postgres/postgres/commit/f696c0cd5f299f1b51e214efc55a22a782cc175d

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -5,7 +5,7 @@ on:
 
 env:
   PIP_BREAK_SYSTEM_PACKAGES: "1"
-  LIBPQ_VERSION: "16.0"
+  LIBPQ_VERSION: "17.4"
   OPENSSL_VERSION: "1.1.1w"
 
 jobs:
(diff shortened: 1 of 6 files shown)


dvarrazzo:
You are misguided: the change is in the database server and the consequence of the change is that a SQL statement that previously used to work doesn't work anymore.

This is independent from the version of the libpq library, which only affects the communication between client and server. In order to communicate with PostgreSQL 17 any supported libpq works and bumping to v17 will not fix your error.

ngnpope:
Urgh. Yes. I didn't look closely enough and this was due to a separate call to `psql ... --command='\\l'`. Sorry for the noise.

PR #1813, opened 2025-10-28
Add riscv64 support for linux builds

mliezun:
Hi!

Here's a patch that builds `riscv64` binaries in Github actions.

The binary distribution facilitates the installation of `psycopg2` as it avoids having to compile it each time that it gets installed.

Thank you for your review!

### DIFF
--- a/.github/workflows/build-and-cache-libpq.yml
+++ b/.github/workflows/build-and-cache-libpq.yml
@@ -1,0 +1,138 @@
+name: Build and cache libpq
+
+# Build the libpq package and cache the artifacts.
+#
+# Every Python version on the same architecture will use the same libpq.
+# Therefore building and caching the libpq together with the binary packages
+# result in multiple concurrent builds, and the github artifacts manageer very
+# confused.
+#
+# This job builds the libpq and then caches the artifacts so that the
+# packages.yml workflow will find the library in the cache.
+#
+# You can see the caches at https://github.com/psycopg/psycopg2/actions/caches
+#
+# Or from the API:
+#
+#   curl -fsSL -X GET \
+#     -H "Accept: application/vnd.github+json" \
+#     -H "Authorization: Bearer $GITHUB_TOKEN" \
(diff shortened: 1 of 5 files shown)


mliezun:
> Do you want to propose a similar MR for psycopg 3 or should I do it myself? The build script is pretty much the same (but with the care to run it only once for every Python release).

Thanks for your review @dvarrazzo.

I'll create a new PR for psycopg 3 😉

.github/workflows/build-and-cache-libpq.yml
```diff
@@ -0,0 +1,137 @@
...
+
+# Build the libpq package and cache the artifacts.
+#
+# Every Python version on the same architecture will use the same libpq.
+# Therefore building and caching the libpq together with the binary packages
+# result in multiple concurrent builds, and the github artifacts manageer very
+# confused.
+#
+# This job builds the libpq and then caches the artifacts so that the
+# packages-bin.yml workflow will find the library in the cache.
```
dvarrazzo:
```suggestion
# packages.yml workflow will find the library in the cache.
```

mliezun:
> Hello,
> 
> the PR seems in good shape, but I don't see a package workflow completed. Please let me know if you manage to make one.
> 
> Also please squash the branch in a single commit or in a few meaningful one: the way you did it the PG3 PR is perfect, including the commit messages 💜
> 
> Let me know when we are ready to merge. I will rebuild the packages on the official side and can release them immediately (for psycopg 3 they will be released with 3.3.0, which should happen soon).

Hi Daniele!

I squashed the PR into a single commit ✅ 

Also, the build packages workflow has completed successfully, see here: https://github.com/mliezun/psycopg2/actions/runs/18902919058

I think this is ready to merge 🚀

dvarrazzo:
Looks good! Let's merge.

dvarrazzo:
riscv64 packages released :+1:

wareagle461 (approved):

PR #1828, opened 2026-02-26
fix: replace 1 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


dvarrazzo:
The commit is a `docs`, not a `fix`. I have fixed the commit message. Thank you very much!

PR #1836, opened 2026-04-16
fix: overflow in malformed interval

### LINKED ISSUE #1835: CWE-190, CWE-835 psycopg2 INTERVAL typecaster infinite loop / client-side DoS
## Summary

`psycopg2` can enter an infinite loop while typecasting a text-format
PostgreSQL `interval` value with a long fractional-second component.

Relevant weakness classes:

- CWE-190: Integer Overflow or Wraparound
- CWE-835: Loop with Unreachable Exit Condition

I reproduced this on `psycopg2-binary==2.9.11`.

The minimal trigger is:

```python
import psycopg2.extensions as ext

ext.INTERVAL(b"0:0:0." + b"0" * 31, None)
```

The call does not return and thread/core consumption spikes to 100%.

## Root cause

The issue is in `typecast_PYINTERVAL_cast()` in
`psycopg/typecast_datetime.c`.

The parser tracks the denominator for fractional seconds in a signed `int` named `denom`, initialized to `1`.

For each digit after the decimal point, it executes:

```c
denom *= 10;
```

For a fractional component of 31 zero digits:

1. The fractional value accumulator remains `0`, so the existing value overflow check does not fire.
2. `denom` overflows the signed `int`.
3. The later normalization loop checks `denom < 1000000L`.
4. The overflowed denominator remains below that threshold and can become `0`.
5. The loop condition never becomes false.

The observed result is an infinite loop and CPU burn.

## Suggested fix

Do not allow `denom` to grow unbounded as a signed `int`.

The parser only needs microsecond precision, so it should cap denominator
tracking at a small bounded value, ignore or round additional fractional digits,
or reject overlong fractional components cleanly before overflow is possible.

One minimal patch is to keep one extra fractional digit for rounding, then ignore
additional fractional digits. This keeps `denom` bounded at `10000000` and avoids
the overflow path:

```diff
diff --git a/psycopg/typecast_datetime.c b/psycopg/typecast_datetime.c
index 0000000..0000000 100644
--- a/psycopg/typecast_datetime.c
+++ b/psycopg/typecast_datetime.c
@@ -318,7 +318,8 @@ typecast_PYINTERVAL_cast(const char *str, Py_ssize_t len, PyObject *curs)
 {
     long v = 0, years = 0, months = 0, hours = 0, minutes = 0, micros = 0;
     PY_LONG_LONG days = 0, seconds = 0;
-    int sign = 1, denom = 1, part = 0;
+    int sign = 1, part = 0;
+    long denom = 1;
     const char *orig = str;
 
     if (str == NULL) { Py_RETURN_NONE; }
@@ -335,6 +336,12 @@ typecast_PYINTERVAL_cast(const char *str, Py_ssize_t len, PyObject *curs)
         case '5': case '6': case '7': case '8': case '9':
             {
                 long v1;
+                /*
+                 * Keep at most seven fractional second digits: six for
+                 * microsecond precision plus one guard digit for rounding.
+                 */
+                if (part == 6 && denom >= 10000000L) { break; }
+
                 v1 = v * 10 + (*str - '0');
                 /* detect either a rollover, happening if v is really too short,
                  * or too big value. On Win where long == int the 2nd check
@@ -358,7 +365,7 @@ typecast_PYINTERVAL_cast(const char *str, Py_ssize_t len, PyObject *curs)
                 v = v1;
             }
-            if (part == 6) {
+            if (part == 6 && denom < 10000000L) {
                 denom *= 10;
             }
             break;
```

## Impact

This is a client-side denial of service. A process or worker using psycopg2 can
hang while fetching a result row.

One remote scenario is a BYODB/SaaS integration:

1. A SaaS product accepts a customer-supplied PostgreSQL DSN. (Bring Your Own DB)
2. The attacker supplies a PostgreSQL-compatible endpoint they control.
3. The victim service connects with psycopg2 and runs a health-check,
   introspection, or normal query.
4. The attacker-controlled endpoint returns a result column marked as
   PostgreSQL `interval` OID `1186` in text format.
5. The returned field value is `0:0:0.` followed by at least 31 fractional-second digits.
6. psycopg2 dispatches to the `INTERVAL` typecaster and the worker hangs.

```mermaid
sequenceDiagram
    participant A as Attacker
    participant S as Victim SaaS using psycopg2
    participant P as Attacker-controlled PostgreSQL endpoint
    A->>S: Submit BYODB DSN
    S->>P: Connect with psycopg2
    P-->>S: PostgreSQL startup/auth OK
    S->>P: Query
    P-->>S: RowDescription: OID 1186 interval, text format
    P-->>S: DataRow: "0:0:0." + 31 fractional digits
    S->>S: INTERVAL typecaster infinite loop
```

Happy to provide the POC for this scenario with a PostgreSQL-compatible python server and psycopg2 client.

dvarrazzo:
Close #1835

@michaelknap does it seem ok?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,12 @@
+Future releases
+---------------
+
+What's new in psycopg 2.9.12 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix infinite loop with malformed interval (:ticket:`1835`).
+
+
 Current release
 ---------------
 
--- a/psycopg/typecast_datetime.c
+++ b/psycopg/typecast_datetime.c
@@ -358,7 +358,12 @@ typecast_PYINTERVAL_cast(const char *str, Py_ssize_t len, PyObject *curs)
                 v = v1;
             }
             if (part == 6) {
-                denom *= 10;
+                /* if denom would overflow it means that an excess of decimal
+                 * digits is being pushed. Postgres should never send more than
+                 * 6. Therefore we can quietly discard the extra. */
+                if (denom <= INT_MAX / 10) {
+                    denom *= 10;
+                }
             }
             break;
 
--- a/tests/test_dates.py
+++ b/tests/test_dates.py
@@ -479,6 +479,12 @@ class DatetimeTests(ConnectingTestCase, CommonDatetimeTestsMixin):
         cur.execute("select '1 day 2 hours'::interval")
         self.assertRaises(psycopg2.NotSupportedError, cur.fetchone)
 
+    def test_bug_1835(self):
+        for digits in (8, 31, 100):
+            self.assertEqual(
+                psycopg2.extensions.INTERVAL(b"0:0:0." + b"0" * digits, None),
+                timedelta(0))
+
 
 class FromTicksTestCase(unittest.TestCase):
     # bug "TimestampFromTicks() throws ValueError (2-2.0.14)"


michaelknap (changes requested):
Thanks for the quick fix! I think the current patch fixes the original <31-digit
hang, but it still leaves an integer overflow in `denom`.

The patch keeps `denom` as a signed `int`:

```c
int sign = 1, denom = 1, denom_digits = 0, part = 0;
```

and rejects only after more than 15 fractional digits:

```c
if (++denom_digits > 15) {
    PyErr_SetString(DataError, "too many fractional digits");
    return NULL;
}
denom *= 10;
```

On common platforms where `int` is 32-bit, `denom *= 10` overflows at the 10th
fractional digit:

```text
1 digit  -> 10
6 digits -> 1000000
9 digits -> 1000000000
10 digits -> 10000000000  // overflows 32-bit int
```

So inputs with roughly 10-15 fractional digits can still execute signed integer
overflow before the new `DataError` path is reached. That may no longer trigger
the original infinite loop, but it still leaves undefined behavior and can
produce incorrect interval parsing.

A simple test case for this remaining issue is:

```python
import psycopg2.extensions as ext

ext.INTERVAL(b"0:0:0." + b"0" * 12, None)
```

I would suggest guarding before multiplication, for example:

```c
if (part == 6) {
    if (denom > INT_MAX / 10) {
        PyErr_SetString(DataError, "too many fractional digits");
        return NULL;
    }
    denom *= 10;
}
```

Alternatively, if the intended behavior is to allow up to 15 fractional digits,
`denom` should be widened to a 64-bit type such as `PY_LONG_LONG`, and the code
should still guard before multiplying. 

Hope this helps! :)

dvarrazzo:
Hmm, you are right, I picked 15 to "stay into a 16 bits signed int" but those are binary digits, not decimal digits.

Probably better to detect the overflow in the loop. I dropped exception raising. Dropping digits is enough.

michaelknap:
> Hmm, you are right, I picked 15 to "stay into a 16 bits signed int" but those are binary digits, not decimal digits.
> 
> Probably better to detect the overflow in the loop. I dropped exception raising. Dropping digits is enough.

Thank you for quick updates and I'm sorry for my late reply. 

I think quietly truncating extra fractional digits is a reasonable behavior, but
this version still has a signed integer overflow issue.

The overflow happens here before the comparison can detect it:

```c
int new_denom = denom * 10;
```

Because `denom` is a signed `int`, `denom * 10` has undefined behavior once the
mathematical result exceeds `INT_MAX`. Checking `new_denom >= denom` after the
multiplication is too late.

The comparison also does not reliably detect wraparound. For example, on common
32-bit two's-complement behavior:

```text
1000000000 * 10 -> 1410065408
```

The wrapped value is still greater than the old denominator, so the overflowed
value could be accepted.

If the desired behavior is to quietly truncate extra fractional digits, I think
the multiplication should be guarded before it happens:

```c
if (part == 6) {
    if (denom <= INT_MAX / 10) {
        denom *= 10;
    }
}
```

This may require including `<limits.h> for INT_MAX.

dvarrazzo:
@michaelknap you are right: I had assumed a two's complement but the behaviour is defined to be... undefined.

So checking for overflow as you suggest seems the right call.

Pushed and rebased.

michaelknap (approved):
Thank you for the changes and for all the work that goes into this module :) All the best.

PR #1839, opened 2026-04-22
Revise security policy for supported versions

0x5t4l1n:
Updated the security policy to reflect supported versions and the reporting process. Vulnerabilities should be reported via GitHub Security Advisory so they can remain private and sensitive data stays protected.


dvarrazzo:
not interested, thank you.

